Issue #16, opened 2018-01-03
Implement persistent queueing in crawler


geneh:
Change CRAWLER_QUEUE_PROVIDER from in-memory to Azure Service Bus.

PR #20, opened 2018-01-04
Configure service bus using AMQP 1.0

geneh:


geneh:
https://github.com/clearlydefined/crawler/issues/16

Issue #9, opened 2018-01-02
log when data types cannot be processed


jeffmcaffer:
In particular we want to know when we are crawling things that are not understood (e.g., source providers we don't know) so we can prioritize work)

jeffmcaffer:
this is different from #24. There are lots of places where a provider cannot be found.

PR #24, opened 2018-01-05
Error out if an unknown provider is found

geneh:
https://github.com/clearlydefined/crawler/issues/9

### DIFF
--- a/lib/sourceDiscovery.js
+++ b/lib/sourceDiscovery.js
@@ -35,8 +35,7 @@ module.exports = async function searchForRevisions(version, candidateUrls, optio
 }
 
 async function discoverRevision(version, candidate, options) {
-  // const provider = getProvider(candidate)
-  const provider = 'github';
+  const provider = getProvider(candidate);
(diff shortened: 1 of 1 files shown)


lib/sourceDiscovery.js
```diff
@@ -91,10 +90,11 @@ async function discoverFromGitHub(version, candidate, options) {
...
-  const lowerUrl = url.toLowerCase();
-  if (lowerUrl.indexOf('github.com'))
+function getProvider(location) {
+  const lowerUrl = location && location.href ? location.href.toLowerCase() : '';
+  if (lowerUrl.includes('github.com')) {
     return 'github';
+  }
   // TODO add other source location extractors
-  return null
+  throw new Error(`Provider could not be found for source at ${lowerUrl}`);
```
jeffmcaffer: suggest a little more context in the error. E.g., `SourceDiscovery provider could not be found for ${lowerUrl}`. There will be other cases where providers cannot be found.

Issue #39, opened 2018-01-10
Create scancode build definition for large repos and run it from crawler


geneh:
Manually create a VSTS build task that would:

- accept a repo with commit hash and a callback command.
- call scancode Docker image to run scancode on a specific repo.
- define Error Checker task to check whether the build is successful or not.
- Publish/drop the scancode JSON results.
- Call the callback command with the scancode results location.

Call this build definition with the required build parameters from crawler for large repos.

PR #41, opened 2018-01-12
Add VSTS build for scanning large repos using scancode

geneh:
Issue https://github.com/clearlydefined/crawler/issues/39

### DIFF
--- a/providers/process/scancode.js
+++ b/providers/process/scancode.js
@@ -40,6 +40,7 @@ class ScanCodeProcessor extends BaseHandler {
         await vstsBuild.queueBuild(document, spec, request.url);
       } catch (error) {
         this.logger.error(error, `${request.toString()} - error queueing build`);
+        return request.markRequeue('VSTS', 'Error queueing build');
       }
       return request.markNoSave();
     }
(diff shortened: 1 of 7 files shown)


providers/process/scancode.js
```diff
@@ -32,11 +33,20 @@ class ScanCodeProcessor extends BaseHandler {
 
   async handle(request) {
     const { document, spec } = super._process(request);
+    if (document.size && document.size > this.options.maxSize) {
+      this.logger.info(`Analyzing ${request.toString()} using ScanCode in VSTS build. Repo size: ${document.size} KB.`);
+      try {
+        const vstsBuild = new VstsBuild(this.options.build);
+        await vstsBuild.queueBuild(document, spec, request.url);
+      } catch (error) {
+        this.logger.error(error, `${request.toString()} - error queueing build`);
```
jeffmcaffer: we likely want to retry this if it was just an error queueing. that may be transient. Need to markRequeue I think it is

providers/process/scancode.js
```diff
@@ -32,11 +33,20 @@ class ScanCodeProcessor extends BaseHandler {
 
   async handle(request) {
     const { document, spec } = super._process(request);
+    if (document.size && document.size > this.options.maxSize) {
+      this.logger.info(`Analyzing ${request.toString()} using ScanCode in VSTS build. Repo size: ${document.size} KB.`);
```
jeffmcaffer: nice. might be interesting to add a logging object that has the size in it (and any thing else) so we can do metrics/analysis around size and time.

Issue #37, opened 2018-01-10
Handle scancode callback request from scancode build


geneh:
We'd like to be able to call an API from scancode build to either ack or nack the message on a queue when a build is either successful or not.
The queueing web API needs to have a new type to handle that.

PR #43, opened 2018-01-16
Gh/ingest vsts

geneh:
Address
https://github.com/clearlydefined/crawler/issues/37
https://github.com/clearlydefined/crawler/issues/42

### DIFF
--- a/lib/build/vsts.js
+++ b/lib/build/vsts.js
@@ -18,13 +18,13 @@ class VstsBuild {
 
   async queueBuild(document, spec, requestUrl) {
     const definitions = await this.vstsBuild.getDefinitions(this.options.project, this.options.definitionName);
-    if (!definitions && !definitions[0]) {
-      // TODO: create build definition
-      throw new Error(`No build definition found for ${this.options.definitionName}`);
+    let definitionId = definitions ? definitions[0] : null;
+    if (!definitionId) {
+      definitionId = (await this.vstsBuild.createDefinition(this.buildDefinition, this.options.project)).id;
     }
     const build = {
       definition: {
-        id: definitions[0].id
+        id: definitionId
       },
       parameters: JSON.stringify({
         sourceUrl: `https://github.com/${spec.namespace}/${spec.name}.git`,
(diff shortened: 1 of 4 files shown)


geneh:
Merging for now.

providers/process/vsts.js
```diff
@@ -0,0 +1,67 @@
...
+  get toolSpec() {
+    return { tool: 'scancode', toolVersion: this.schemaVersion };
+  }
+
+  get authToken() {
+    return 'Basic ' + new Buffer(this.options.apiToken + ':', 'utf8').toString('base64');
+  }
+
+  canHandle(request) {
+    _toolVersion = request.document.toolVersion || '2.2.1';
```
jeffmcaffer: this is shared state. There may be many crawler loops running at once and they will all overwrite each other.

providers/process/vsts.js
```diff
@@ -0,0 +1,67 @@
...
+let _toolVersion;
+
+class VstsProcessor extends BaseHandler {
+
+  get schemaVersion() {
+    return _toolVersion;
+  }
+
+  get toolSpec() {
+    return { tool: 'scancode', toolVersion: this.schemaVersion };
```
jeffmcaffer: This should not be scancode specific. If we want to run fossology in a build, this ingest processor should be able to handle it. The request queued into the crawler by the build should carry all the context. (e.g., tool name, version, ...)

providers/process/vsts.js
```diff
@@ -0,0 +1,67 @@
...
+
+  canHandle(request) {
+    _toolVersion = request.document.toolVersion || '2.2.1';
+    return request.type === 'ingest-vsts';
+  }
+
+  async handle(request) {
+    const { document, spec } = super._process(request);
+    this.addBasicToolLinks(request, spec);
+    this.logger.info(`Processing ${request.toString()}`);
```
jeffmcaffer: this will be a little noisy as `info`. Either remove this in favor of a more general approach to "starting" and "completing" processing within the crawler or drop the log level.

providers/process/vsts.js
```diff
@@ -0,0 +1,67 @@
...
+    const file = this._createTempFile(request);
+    await this._getBuildOutput(document.buildOutput, file.name);
+    const dir = this._createTempDir(request);
+    await decompress(file.name, dir.name, { strip: 1 });
+    try {
+      const scancodeFilePath = `${dir.name}${path.sep}scancode.json`;
```
jeffmcaffer: either the file name needs to be generic of the queued request should spec the name. This is specific scancode

lib/build/vsts.js
```diff
@@ -18,13 +18,13 @@ class VstsBuild {
 
   async queueBuild(document, spec, requestUrl) {
     const definitions = await this.vstsBuild.getDefinitions(this.options.project, this.options.definitionName);
-    if (!definitions && !definitions[0]) {
-      // TODO: create build definition
-      throw new Error(`No build definition found for ${this.options.definitionName}`);
+    let definitionId = definitions ? definitions[0] : null;
+    if (!definitionId) {
+      definitionId = (await this.vstsBuild.createDefinition(this.buildDefinition, this.options.project)).id;
```
jeffmcaffer: Factor the first few lines of this method into a separate method that just ensures the build definition exists and returns the id. This will make for easier testing and separates concerns

lib/build/vsts.js
```diff
@@ -41,12 +41,110 @@ class VstsBuild {
       url: requestUrl,
       payload: {
         body: {
+          toolVersion: '__TOOL_VERSION__', // will be replaced by VSTS build
```
jeffmcaffer: might need some more info here like the tool name and the name of the file containing the output (unless that will be either canonicalized or the info is to be pulled from the file name itself)

PR #58, opened 2018-02-18
update VSTS build integration

jeffmcaffer:
When we run scans as VSTS builds this PR
* makes the build output more self-described so the outputs can simply be downloaded and ingested (if needed)
* generalizes the way "size" is computed to determine if the build approach should be used. Rather than using git repo size, we now look on disk at the number of files and the size of the files. I've made guesses at what makes sense to run inline vs build but we'll have to see that over time.

### DIFF
--- a/providers/fetch/mavenFetch.js
+++ b/providers/fetch/mavenFetch.js
@@ -43,20 +43,16 @@ class MavenFetch extends BaseHandler {
   }
 
   async _getArtifact(spec, destination) {
-    const extension = spec.type === 'sourcearchive' ? '-sources.jar' : '.pom';
-    const uri = this._buildUrl(spec, extension);
-    var options = {
-      method: 'GET',
-      uri
-    };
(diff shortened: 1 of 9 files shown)


providers/fetch/mavenFetch.js
```diff
@@ -43,20 +43,16 @@ class MavenFetch extends BaseHandler {
...
         if (error)
           return reject(error);
-        if (response.statusCode === 200 || response.statusCode === 404)
-          return resolve(response.statusCode);
-        reject(new Error(`${response.statusCode} ${response.statusMessage}`))
-      }).pipe(fs.createWriteStream(destination));
+        if (response.statusCode === 404)
+          resolve(response.statusCode);
+        if (response.statusCode !== 200)
+          reject(new Error(`${response.statusCode} ${response.statusMessage}`))
```
geneh: It would be good to run `npm test` here to resolve any linting errors.

Issue #51, opened 2018-02-13
Crawler improvements for maven


fossygirl:
Need a maven expert to take a look at what we've done wrt
* source discovery
* fetching from maven central
* how this might apply to other maven repos
* POM interpretation
* URL construction for artifacts

What we have now works for some number of artifacts but likely has some issues.

geneh:
This is done, closing. 
Thanks, @fernomac!

PR #71, opened 2018-02-28
Adding support for crawling maven parent poms

fernomac:
As an initial cut at helping out with #51, adding support for also spidering in [parent poms](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance) from Maven Central and doing a rough job of merging things together.

For example, [rxgrpc](https://search.maven.org/remotecontent?filepath=com/salesforce/servicelibs/rxgrpc/0.8.0/rxgrpc-0.8.0.pom)'s pom doesn't directly declare SCM info or a license. That stuff is implicitly picked up from [the parent pom](https://search.maven.org/remotecontent?filepath=com/salesforce/servicelibs/reactive-grpc/0.8.0/reactive-grpc-0.8.0.pom) it shares with other Maven submodules in the same repository.

As per offline conversations with @jeffmcaffer this changes the `manifest` to include both a "summary" approximating the actual merged pom as Maven sees it, and the audit trail of actual pom files that was used to assemble the summary. We'll need a corresponding change in the service, which I'll tackle tomorrow.

The merge it's doing currently is pretty naive, ie if the child pom declares a `<properties>` element that will completely override any properties from the parent pom, as opposed to doing a property-by-property merge. That's hopefully not too huge of a concern for anything we're interested in here?


jeffmcaffer:
@geneh can you rebase this and resolve the merge conflicts (likely just file headers). and then integrate and see that it's doing the right things? Would be great to then run some of the top 1K maven packages through (see https://github.com/clearlydefined/service/issues/84). Queuing them up as maven/mavencentral/<group>/<artifact>/<version> should do it.  The hope is that that will get us a fair bit of data as well as both the sources.jar location and the GitHub location.

Issue #81, opened 2018-03-07
Consider upgrading scancode


geneh:
https://github.com/nexB/scancode-toolkit/releases/
Upgrade from version 2.2.1 to 2.9.0b1 mostly to improve performance.
Affected clearlydefined projects: crawler, tool-images, possibly service.

geneh:
There are minor changes in the tool's options, for example: `--diag` has become `--license-diag`; `-f json-pp ./input-dir ./output.json` has become `--json-pp ./output.json ./input-dir`
JSON output is almost the same with a few minor exceptions.

When the following options are used, there is a very significant performance improvement, especially for large repos. Small repos aren't affected as much.

- Relatively small repo (9 sec in both versions):
```
git clone git@github.com:Marak/colors.js.git

SECONDS=0; /Users/gene/Downloads/scancode-toolkit-2.2.1/scancode --copyright --license --info --package --diag --only-findings --strip-root --quiet --timeout 1000 -n 2 -f json-pp ./colors.js/ ./output221-colors.json; cat ./output221-colors.json | wc -l ; echo $SECONDS
     497
9

SECONDS=0; /Users/gene/Downloads/scancode-toolkit-2.9.0b1/scancode --copyright --license --info --package --license-diag --only-findings --strip-root --quiet --timeout 1000 -n 2 --json-pp ./output290-colors.json ./colors.js/; cat ./output290-colors.json | wc -l ; echo $SECONDS
     506
9
```
- Huge repo (859 vs 58 sec!):
```
git clone git@github.com:facebook/react.git

SECONDS=0; /Users/gene/Downloads/scancode-toolkit-2.2.1/scancode --copyright --license --info --package --diag --only-findings --strip-root --quiet --timeout 1000 -n 2 -f json-pp ./react ./output221-react.json; cat ./output221-react.json | wc -l ; echo $SECONDS
   30603
859


SECONDS=0; /Users/gene/Downloads/scancode-toolkit-2.9.0b1/scancode --copyright --license --info --package --license-diag --only-findings --strip-root --quiet --timeout 1000 -n 2 --json-pp ./output290-react.json ./react; cat ./output290-react.json | wc -l ; echo $SECONDS
   31463
58
```

geneh:
If a repo is scanned with version 2.9.0b1, the service cannot determine the latest scancode version because according to semver module this is an invalid version format:
`.reduce((max, current) => (semver.gt(current, max) ? current : max), versions[0])`
[...]

PR #82, opened 2018-03-08
Upgrade scancode to version 2.9.0b1

geneh:
Issue https://github.com/clearlydefined/crawler/issues/81

### DIFF
--- a/lib/baseHandler.js
+++ b/lib/baseHandler.js
@@ -84,7 +84,11 @@ class BaseHandler {
     if (versions.length === 1) return versions[0]
     return versions
       .filter(v => !this.isPreReleaseVersion(v))
-      .reduce((max, current) => (semver.gt(current, max) ? current : max), versions[0])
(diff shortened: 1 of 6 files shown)


lib/baseHandler.js
```diff
@@ -84,7 +84,11 @@ class BaseHandler {
     if (versions.length === 1) return versions[0]
     return versions
       .filter(v => !this.isPreReleaseVersion(v))
-      .reduce((max, current) => (semver.gt(current, max) ? current : max), versions[0])
+      .filter(v => semver.prerelease(v) === null)
+      .reduce(
+        (max, current) => (semver.gt(semver.coerce(current).version, semver.coerce(max).version) ? current : max),
```
jeffmcaffer: modify the code that gets --version to change from 2.9.0b1 to 2.9.0-b1 (or some related valid semver) and remove the coerce.

Here and everywhere that we get the version of the tool.

Issue #89, opened 2018-03-23
EISDIR & absolute path errors


geneh:
The following Maven components have ended up in dead letter queue:
```
1 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.scala-lang/scala-reflect/2.13.0-M3-f73b161' 'EISDIR: illegal operation on a directory, open \'/tmp/cd-Ht1z4x/scala/reflect/\''
2 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.scala-lang/scala-library/2.13.0-M3-f73b161' 'EISDIR: illegal operation on a directory, open \'/tmp/cd-BhEZPG/scala/annotation/\''
3 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.xhtmlrenderer/flying-saucer-pdf-itext5/9.1.12' 'absolute path: /'
4 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.jsonschema2pojo/jsonschema2pojo-cli/1.0.0-alpha2' 'absolute path: /'
5 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.infinispan/infinispan-embedded-query/9.1.7.Final' 'EISDIR: illegal operation on a directory, open \'/tmp/cd-7dkcT0/org/infinispan/\''
6 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.xhtmlrenderer/flying-saucer-swt/9.1.12' 'absolute path: /'
7 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.xhtmlrenderer/flying-saucer-log4j/9.1.12' 'absolute path: /'
8 'clearlydefined' 'cd:/sourcearchive/mavencentral/com.facebook.presto/presto-jdbc/0.197' 'EISDIR: illegal operation on a directory, open \'/tmp/cd-1AWdll/com/facebook/\''
9 'clearlydefined' 'cd:/sourcearchive/mavencentral/org.infinispan/infinispan-embedded/9.1.7.Final' 'EISDIR: illegal operation on a directory, open \'/tmp/cd-GXCWCD/org/infinispan/\''
```

```
17 'scancode' 'cd:/sourcearchive/mavencentral/xerces/xercesImpl/2.11.0' 'EISDIR: illegal operation on a directory, open \'/tmp/cd-y14wKn/org/apache/\''
[...]

PR #91, opened 2018-04-03
Resolve EISDIR error

geneh:
Address issue https://github.com/clearlydefined/crawler/issues/89
This error has been happening mostly for some of Maven Central artifacts with broken links in jar files.
decompress library is still able to decompress those archives when EISDIR error is encountered.

### DIFF
--- a/lib/baseHandler.js
+++ b/lib/baseHandler.js
@@ -71,7 +71,10 @@ class BaseHandler {
   }
 
   decompress(source, destination) {
-    return decompress(source, destination)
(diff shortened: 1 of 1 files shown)


pombredanne:
@geneh Do you have an example URL for a Jar with such a problem?

geneh:
@pombredanne Yes, https://search.maven.org/remotecontent?filepath=com/twitter/parquet-jackson/1.6.0/parquet-jackson-1.6.0-sources.jar
I totally forgot about scancode's ability to decompress archives. It would be nice to try.

PR #93, opened 2018-04-04
normalize NuGet version

geneh:
Address issue https://github.com/clearlydefined/crawler/issues/92


providers/fetch/nugetFetch.js
```diff
@@ -52,6 +52,31 @@ class NuGetFetch extends BaseHandler {
     return body
   }
 
+  // https://docs.microsoft.com/en-us/nuget/reference/package-versioning#normalized-version-numbers
+  _normalizeVersion(version) {
+    // 1.00 is treated as 1.0, 1.01.1 is treated as 1.1.1, 1.00.0.1 is treated as 1.0.0.1
+    let parts = version.split('.')
+    for (let i = 0; i < parts.length; i++) {
```
jeffmcaffer: This should be equivalent.
```
const trimmed = version.split('.').map(part => part.replace(/^0+(?!\.|$)/, '')); 
return (trimmed[3] === '0' ? trimmed.slice(0,3) : trimmed).join('.')
```

geneh: Nice! It looks even better with lodash:
`const trimmed = version.split('.').map(part => _.trimStart(part, '0') || '0')`

PR #94, opened 2018-04-04
NPM name must be lower case

geneh:
Tested with `{ "type": "npm", "url": "cd:/npm/npmjs/-/SlickGrid/2.3.3" }`

### DIFF
--- a/lib/entitySpec.js
+++ b/lib/entitySpec.js
@@ -8,7 +8,7 @@ const NONE = 0
 
 const toLowerCaseMap = {
   github: NAMESPACE | NAME,
-  npmjs: NONE,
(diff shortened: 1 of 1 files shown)


jeffmcaffer:
Can you detail the scenario here? In past investigations NPM was found to be case-sensitive. npmjs.com is. compare 
https://www.npmjs.com/package/SlickGrid
https://www.npmjs.com/package/slickgrid

IIRC there was some talk in the doc about being lower case but in practice (historically) there were packages with mixed case and they were different (or some such).

What problem is this change addressing?

geneh:
Right, so those were bad registrations in our system. The code works as expected.

PR #103, opened 2018-07-05
Add one FOSSology tool nomos to crawler

larainema:
This pull request add the FOSSology tool nomos to crawler, also update the Docker file to package the FOSSology tools.

One issue want to discuss how to handle the output as the output is not json format.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,19 +16,9 @@ ENV SCANCODE_HOME=/opt/scancode-toolkit-2.9.2
 WORKDIR /opt
 RUN git clone https://github.com/fossology/fossology.git
 
-WORKDIR /opt/fossology
-RUN apt-get update && \
-    apt-get install -y lsb-release sudo postgresql php5-curl libpq-dev libdbd-sqlite3-perl libspreadsheet-writeexcel-perl && \
-    /opt/fossology/utils/fo-installdeps -e -y && \
-    rm -rf /var/lib/apt/lists/*
-RUN curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer
-RUN /opt/fossology/install/scripts/install-spdx-tools.sh
-RUN /opt/fossology/install/scripts/install-ninka.sh
-RUN make install
-RUN /etc/init.d/postgresql start
-RUN /usr/local/lib/fossology/fo-postinstall
-ENV FOSSOLOGY_HOME=/usr/local/etc/fossology/mods-enabled
+WORKDIR /opt/fossology/src/nomos/agent
+RUN make -f Makefile.sa
+ENV FOSSOLOGY_HOME=/opt/fossology/src/nomos/agent
 
 COPY package*.json /tmp/
 RUN cd /tmp && npm install --production
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,9 @@ RUN curl -sS https://getcomposer.org/installer | php && \
 RUN /opt/fossology/install/scripts/install-spdx-tools.sh
 RUN /opt/fossology/install/scripts/install-ninka.sh
 RUN make install
-ENV FOSSOLOGY_HOME=/usr/local/share/fossology
+RUN /etc/init.d/postgresql start
+RUN /usr/local/lib/fossology/fo-postinstall
+ENV FOSSOLOGY_HOME=/usr/local/etc/fossology/mods-enabled
 
 COPY package*.json /tmp/
 RUN cd /tmp && npm install --production
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,22 @@ RUN curl -sL https://github.com/nexB/scancode-toolkit/releases/download/v2.9.2/s
   && /opt/scancode-toolkit-2.9.2/scancode --version
 ENV SCANCODE_HOME=/opt/scancode-toolkit-2.9.2
 
+# FOSSology
+WORKDIR /opt
+RUN git clone https://github.com/fossology/fossology.git
+
+WORKDIR /opt/fossology
+RUN apt-get update && \
+    apt-get install -y lsb-release sudo postgresql php5-curl libpq-dev libdbd-sqlite3-perl libspreadsheet-writeexcel-perl && \
+    /opt/fossology/utils/fo-installdeps -e -y && \
+    rm -rf /var/lib/apt/lists/*
+RUN curl -sS https://getcomposer.org/installer | php && \
+    mv composer.phar /usr/local/bin/composer
+RUN /opt/fossology/install/scripts/install-spdx-tools.sh
+RUN /opt/fossology/install/scripts/install-ninka.sh
+RUN make install
+ENV FOSSOLOGY_HOME=/usr/local/share/fossology
+
 COPY package*.json /tmp/
 RUN cd /tmp && npm install --production
 RUN mkdir -p "${APPDIR}" && cp -a /tmp/node_modules "${APPDIR}"
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -76,8 +76,6 @@ module.exports = {
     },
     fossology: {
       installDir: config.get('FOSSOLOGY_HOME'),
-      options: [
-      ]
     },
     source: {},
     top: {
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -74,6 +74,15 @@ module.exports = {
         }
       }
     },
+    fossology: {
+      installDir: config.get('FOSSOLOGY_HOME'),
+      options: [
+      ],
+      build: {
+        crawlerUrl: config.get('CRAWLER_SERVICE_URL') || 'http://localhost:5000',
+        crawlerAuthToken: config.get('CRAWLER_SERVICE_AUTH_TOKEN') || 'secret'
+     }
+    },
     source: {},
     top: {
       githubToken
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -77,11 +77,7 @@ module.exports = {
     fossology: {
       installDir: config.get('FOSSOLOGY_HOME'),
       options: [
-      ],
-      build: {
-        crawlerUrl: config.get('CRAWLER_SERVICE_URL') || 'http://localhost:5000',
-        crawlerAuthToken: config.get('CRAWLER_SERVICE_AUTH_TOKEN') || 'secret'
-     }
+      ]
     },
     source: {},
     top: {
--- a/providers/process/fossology.js
+++ b/providers/process/fossology.js
@@ -43,16 +43,15 @@ class FossologyProcessor extends BaseHandler {
 
     return new Promise((resolve, reject) => {
       // TODO add correct parameters and command line here
-      const parameters = [file.name, request.document.location].join(' ')
-      exec(`cd ${this.options.installDir} && .${path.sep}fossology ${parameters}`, (error, stdout, stderr) => {
-        if (this._isRealError(error) || this._hasRealErrors(file.name)) {
+      const parameters = ['-ld', request.document.location, '>', file.name].join(' ')
+      exec(`cd ${this.options.installDir} && .${path.sep}nomos${path.sep}agent${path.sep}nomos ${parameters}`, (error, stdout, stderr) => {
+        if (error) {
           request.markDead('Error', error ? error.message : 'FOSSology run failed')
           return reject(error)
         }
-
         // TODO update to indicate the correct content type for the FOSSology output
         document._metadata.contentLocation = file.name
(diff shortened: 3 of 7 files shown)


Dockerfile
```diff
@@ -12,6 +12,24 @@ RUN curl -sL https://github.com/nexB/scancode-toolkit/releases/download/v2.9.2/s
...
+    rm -rf /var/lib/apt/lists/*
+RUN curl -sS https://getcomposer.org/installer | php && \
+    mv composer.phar /usr/local/bin/composer
+RUN /opt/fossology/install/scripts/install-spdx-tools.sh
+RUN /opt/fossology/install/scripts/install-ninka.sh
+RUN make install
+RUN /etc/init.d/postgresql start
+RUN /usr/local/lib/fossology/fo-postinstall
+ENV FOSSOLOGY_HOME=/usr/local/etc/fossology/mods-enabled
+
```
craigez: Could you use the ready available fossology docker containers instead?

providers/process/npmExtract.js
```diff
@@ -34,6 +34,7 @@ class NpmExtract extends BaseHandler {
       await this._createDocument(request, manifest, request.document.registryData)
     }
     this.linkAndQueueTool(request, 'scancode')
+    this.linkAndQueueTool(request, 'fossology')
```
craigez: Should we call it fossology if it is just the noms agent?

Dockerfile
larainema: Here we didn't want to setup a full FOSSology instance and run FOSSology from scheduler, just want to setup the FOSSology command line tools.

providers/process/npmExtract.js
larainema: This is just a first step trial, will discuss if write one Processor that spawns all the different FOSSology tools or write a different Processor for each FOSSology tool.

craigez: The agents search for very different things. In other projects we've used the both separately and combined. For instance combining the nomos and monk results, but treating the copyright agent results separately.

Dockerfile
craigez: Make sense. Worth considering bring fossology up in a separate container though and then appropriately link the containers together.

jeffmcaffer: Agreed, Lets set up something that automatically builds and publishes a nomossa container. We don't need to build it every time we build the crawler and we should be fixing on a version etc. Note as well that eventually we may want to move to alpine rather than whatever comes with Node. Then we would have to deal with the toolchain needed to build nomos.

@mcjaeger, what do you think about having FOSSology publish a nomossa container? Or should we publish one on dockerhub in the `clearlydefined` namespace.

providers/process/fossology.js
```diff
@@ -41,20 +42,75 @@ class FossologyProcessor extends BaseHandler {
...
+          return reject(error)
+        }
+        resolve(stdout)
+      })
+    })
+  }
+
+  async _getFilelist(document) {
+    return new Promise((resolve, reject) => {
+      exec(`find ${document.location} -type f -print |tr "\\n" " "`, (error, stdout, stderr) => {
```
jeffmcaffer: This only works on linux/mac.  Consider using something like https://www.npmjs.com/package/node-dir

providers/process/fossology.js
```diff
@@ -81,9 +137,10 @@ class FossologyProcessor extends BaseHandler {
   _detectVersion() {
```
jeffmcaffer: Here we will need to take another approach as we are combining multiple agents to make one logical tool with one output. I think we will have to set this at build time from a constant in the source and ensure that the dockerfile (for example) is pulling in particular versions of the agents. As a practical matter, we should be using the nomos and monk and copyright that come from ONE FOSSology version. As such, we can just use the FOSSology version for our tool version.

We may need to refine this approach in the future but since people already know what FOSSology is and may be following along with its versioning scheme, it is best if we do the same. 

One note: tool versions follow the https://semver.org pattern. If FOSSology varies from that, we'll have to fudge it a bit.

Having said all the above, it would be good if the output section for each agent includes the version of the agent.

providers/process/npmExtract.js
jeffmcaffer: It's a good point @craigez . For now at least we have decided to run them all at once. There are a few practical matters (feel free to push for a different direction)

* Folks appear to know more what FOSSology is than they do the individual agents
* ClearlyDefined aggregates the agent outputs anyway. ScanCode's one output includes license, copyright, ... info all in one file. ClearlyDefined teases the interesting info out and structures it as definitions. We can do the same here.
* The individual agent outputs will still be available. The "FOSSology" ClearlyDefined tool output would have N different discrete sections, one for each agent that was run.
* Save on clone time. Since each agent needs all these source, and each ClearlyDefined tool runs independently, structuring each agent as a separate tool would force separate clones.
* Save on blob count and overall management counts. Since each tool has its own output, structuring the agents as separate tools would result in many more blobs. Not in and of itself a problem, just adds to the overall bulk for relatively little gain.  We are running them all anyway.

providers/process/fossology.js
```diff
@@ -41,20 +42,75 @@ class FossologyProcessor extends BaseHandler {
...
+    const file_list = await this._getFilelist(document)
+    // get nomos output
+    const nomosStdout = await this._getNomos(request)
+    // get copyright output
+    //const copyrightStdout = await this._getCopyright(request, file_list)
+    // get monk output
+    //const monkStdout = await this._getMonk(request, file_list)
+
+    console.log(nomosStdout)
+    await writeFile(file.name, nomosStdout)
```
jeffmcaffer: Given that we will eventually have multiple agents and their corresponding output, the document that results from running our FOSSology tool should be structures something like this

```javascript
{
  nomos: {
    version: '1.3.4',
    parameters: ' < this list of parameters used to run the agent >',
    output: {
      contentType: 'text/plain',
      content: ' < base 64 encoding > '
    }
  },
  monk: {
    version: '3.4.5',
    parameters: ' < this list of parameters used to run the agent >',
    output: {
      contentType: 'application/json',
      content: {
        // actual output object here.
      }
    }
  }
}
```

Ideally then, each of `runNomos`, `runMonk` and `runCopyright` would just return its part of the output since it can figure out the version of the tool, format of the output, ...

Given the above structure, code becomes something like 

```javascript
  const document = {
    nomos: runNomos(),
    monk: runMonk(),
    copyright: runCopyright()
  }
  document._metadata = request.document._metadata
  request.document = document
```

Note there is no error handling. We should decide on a policy for errors. Off the top of my head I'd say if a tool errors then log the error (more on that later) and skip it in the output. If all the tools fail then there should be no output (e.g., `request.markNoSave()`)

providers/process/npmExtract.js
craigez: So long as we can categorize different agents or results such that we can easily pull out the different aspects and handle them appropriately we should store it all as FOSSology.

config/cdConfig.js
```diff
@@ -74,6 +74,11 @@ module.exports = {
         }
       }
     },
+    fossology: {
+      installDir: config.get('FOSSOLOGY_HOME'),
+      options: [
```
jeffmcaffer: can you remove options completely?

providers/process/fossology.js
```diff
@@ -36,25 +38,62 @@ class FossologyProcessor extends BaseHandler {
...
+        if (error) {
           request.markDead('Error', error ? error.message : 'FOSSology run failed')
           return reject(error)
         }
+        let buff = new Buffer(stdout)
+        const nomosOutput = {
+          version: this.schemaVersion,
+          parameters: parameters,
+          output: {
+            contentType: 'text/plain',
```
jeffmcaffer: base64 should likely be `application/base64`

providers/process/fossology.js
```diff
@@ -36,25 +38,62 @@ class FossologyProcessor extends BaseHandler {
...
-        if (this._isRealError(error) || this._hasRealErrors(file.name)) {
+      const parameters = ['-ld', request.document.location].join(' ')
+      exec(`cd ${this.options.installDir} && ./nomossa ${parameters}`, (error, stdout, stderr) => {
+        if (error) {
           request.markDead('Error', error ? error.message : 'FOSSology run failed')
           return reject(error)
         }
+        let buff = new Buffer(stdout)
+        const nomosOutput = {
+          version: this.schemaVersion,
```
jeffmcaffer: Here the `version` should be the version of Nomos itself. The overall tool version would be the version FOSSology from which nomos was taken (or something close to that). In particular, it is likely that the version prop for Monk will be different from this one for Nomos.

providers/process/fossology.js
```diff
@@ -36,25 +38,62 @@ class FossologyProcessor extends BaseHandler {
...
+          parameters: parameters,
+          output: {
+            contentType: 'text/plain',
+            content: buff.toString('base64')
+          }
+        }
+        resolve(JSON.parse(JSON.stringify(nomosOutput)))
```
jeffmcaffer: Why stringify and parse? should just be able to resolve `nomosOutput`

providers/process/fossology.js
```diff
@@ -36,25 +38,62 @@ class FossologyProcessor extends BaseHandler {
...
+  }
 
-        // TODO update to indicate the correct content type for the FOSSology output
-        document._metadata.contentLocation = file.name
-        document._metadata.contentType = 'application/json'
-        document._metadata.releaseDate = request.document.releaseDate
-        resolve(request)
+  //TODO: will revisit after FOSSology have copyright standalone version
+  async _runCopyright(request, files) {
+    return new Promise((resolve, reject) => {
```
jeffmcaffer: You might want to either remove these or sketch them out to return a structure more like the Nomos runner. I would tend to say we should just remove them. This is not hard code and getting stand alone Monk and Copyright is not on our immediate horizon. They are easy to add back. We have the structure in place in the result to handle more results.

providers/process/fossology.js
larainema: From my understanding the cut down version of nomos '3.3.0-54-g6ce64cd' is a valid prerelease tag in SemVer, I think just replace the second hyphen with the correct separator characters to make it compatible.
https://github.com/semver/semver/issues/394

Dockerfile
```diff
@@ -12,6 +12,14 @@ RUN curl -sL https://github.com/nexB/scancode-toolkit/releases/download/v2.9.2/s
   && /opt/scancode-toolkit-2.9.2/scancode --version
 ENV SCANCODE_HOME=/opt/scancode-toolkit-2.9.2
 
+# FOSSology
+WORKDIR /opt
+RUN git clone https://github.com/fossology/fossology.git
+
```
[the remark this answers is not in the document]
jeffmcaffer: We can leave this as is for our normal production use but I wonder if there is a way to get nomos compiled for windows and mac so that people can run it for local testing? 

Let's not block on this but it would be great to make it easier in the furture/

providers/process/fossology.js
```diff
@@ -36,25 +44,32 @@ class FossologyProcessor extends BaseHandler {
     const size = await this._computeSize(document)
```
jeffmcaffer: Here we should check that there nomosVersion. In local dev setups there may not be yet the system should continue to work . Suggest just something like

```
return request.markSkip('No nomos tool found')
```

providers/process/fossology.js
```diff
@@ -81,13 +96,35 @@ class FossologyProcessor extends BaseHandler {
   _detectVersion() {
     if (_toolVersion) return _toolVersion
     return new Promise((resolve, reject) => {
-      exec(`cd ${this.options.installDir} && .${path.sep}fossology --version`, (error, stdout, stderr) => {
+      exec(`cd ${this.options.installDir} && ./nomossa -V`, (error, stdout, stderr) => {
```
jeffmcaffer: Should not duplicate the code here. Just reuse the nomos version as the tool version.

providers/process/fossology.js
```diff
@@ -81,13 +96,35 @@ class FossologyProcessor extends BaseHandler {
...
   }
+
+  _detectNomosVersion() {
+    if (_nomosVersion) return _nomosVersion
+    return new Promise((resolve, reject) => {
+      exec(`cd ${this.options.installDir} && ./nomossa -V`, (error, stdout, stderr) => {
+        if (error) return reject(error)
+        _nomosVersion = stdout.replace('nomos\ build\ version:', '').trim()
+        _nomosVersion = _nomosVersion.replace(/r\(.*\)./, '').trim()
+        resolve(_nomosVersion)
```
jeffmcaffer: did you confirm that the nomos version will come out as a semver? It must come as or be made to be a semver

Dockerfile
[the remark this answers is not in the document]
larainema: I think mac should be easy to support, but as I know fossology didn't support windows.

PR #109, opened 2018-07-09
Ruby gem support

Teju-Manchenella:
added ruby gem support


jeffmcaffer:
One thing that @geneh and I noticed is captured in https://github.com/clearlydefined/service/issues/157. We can do that change separately but it gives some context for the change needed here. We need to figure out if Ruby is case sensitive and if so, what parts.

PR #110, opened 2018-07-12
Pypi support

Teju-Manchenella:
Fetch and extract pypi packages. Commented out top function for pypi because it does not use explicit pypi api. It was used for local testing.


providers/process/pypiExtract.js
```diff
@@ -0,0 +1,57 @@
...
+    if (!registryData) {
+      return null
+    }
+    registryData.info.bugtrack_url && candidates.push(registryData.info.bugtrack_url)
+    registryData.info.docs_url && candidates.push(registryData.info.docs_url)
+    registryData.info.download_url && candidates.push(registryData.info.download_url)
+    registryData.info.home_page && candidates.push(registryData.info.home_page)
+    registryData.info.package_url && candidates.push(registryData.info.package_url)
+    registryData.info.project_url && candidates.push(registryData.info.project_url)
+    registryData.info.release_url && candidates.push(registryData.info.release_url)
```
craigez: I'm not 100% versed on how the crawler works, but do we want to also grab the license, summary, etc... ?

Teju-Manchenella: Working on grabbing the license in another function. This function only looks at urls and tries to determine source location.

craigez: For my future reference, would we also get that metadata as part of the crawler?

Teju-Manchenella: Yes, that would occur in the crawler. The crawlers job is to get all the information it can possibly get about a specific package.

PR #118, opened 2018-07-16
Formate licenses for PyPi packages

Teju-Manchenella:
Attempt to find the 'License :: OSI Approved' trove classifier and use spdx-corrector to figure out properly formatted spdx license.

### DIFF
--- a/providers/fetch/pypiFetch.js
+++ b/providers/fetch/pypiFetch.js
@@ -5,6 +5,7 @@ const BaseHandler = require('../../lib/baseHandler')
 const requestRetry = require('requestretry').defaults({ maxAttempts: 3, fullResponse: true })
 const nodeRequest = require('request')
 const fs = require('fs')
+const spdxCorrect = require('spdx-correct')
 const { findLastKey, get, find } = require('lodash')
 
 const providerMap = {
@@ -46,7 +47,8 @@ class PyPiFetch extends BaseHandler {
 
   _createDocument(dir, spec, registryData) {
     const releaseDate = this._extractReleaseDate(spec, registryData)
-    return { location: dir.name, registryData, releaseDate }
+    const declaredLicense = this._extractDeclaredLicense(registryData)
+    return { location: dir.name, registryData, releaseDate, declaredLicense }
   }
 
   _extractReleaseDate(spec, registryData) {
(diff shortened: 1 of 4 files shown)


providers/fetch/pypiFetch.js
```diff
@@ -58,6 +60,19 @@ class PyPiFetch extends BaseHandler {
     return release.upload_time
   }
 
+  _extractDeclaredLicense(registryData) {
+    const classifiers = get(registryData, 'info.classifiers')
+    if (!classifiers) return null
+    for (const classifier in classifiers) {
+      if (classifiers[classifier].includes('License :: OSI Approved ::')) {
```
craigez: SPDX has a wider set of licenses than those that are OSI approved. I don't know, but there could be python packages in PyPi with licenses that are no OSI approved but many people consider to be open source and are present in SPDX.

providers/process/top.js
```diff
@@ -36,6 +36,8 @@ class TopProcessor extends BaseHandler {
         return this._processTopNuGets(request)
       case 'github':
         return this._processAllGitHubOrgRepos(request)
+      // case 'pypi':
+      //   return this._processTopPyPis(request)
```
craigez: Did you mean to include this in this PR?

providers/fetch/pypiFetch.js
Teju-Manchenella: The metadata of a specific pypi package is fetched and stored in the registryData variable. In the metadata there is a path to a set of classifiers ('info.classifiers'). These classifiers are called trove classifiers and are structured in a very specific way. There is a list of predefined trove classifiers that are used to categorize each release, describing who it's for, what systems it can run on, and how mature it is (https://pypi.org/classifiers). This is the only documentation I could find about potential locations for declared licenses in pypi packages. 

I agree that there exist many other licenses; however, there is no documented location or structure of where these other licenses might exist. Due to this fact I only look for that specific classifier when computing the declared licenses. If a license is not able to be found it can always be added by a curator.

craigez: Looks like everything not in that list goes in "Other" then. Agree that anything outside that list you link to will need to be curated.

Teju-Manchenella:
@jamesiri, @jeffmcaffer says that you might be interested in the PR. I am trying to extract all the information from pypi packages and would like some feedback.

PR #121, opened 2018-08-01
proper casing for each package

Teju-Manchenella:
New toLowerCase map across website, service, and crawler.  Addresses issue #[237](https://github.com/clearlydefined/service/issues/157)


dabutvin:
merged into #132

PR #125, opened 2018-08-12
Add ability to store interesting file content

jeffmcaffer:
When crawling components we are interested in capturing the content of key files such as LICENSE, NOTICES, ... and storing them for future use in operations like generating product notice files. This PR adds this ability in the following way:

* Each ClearlyDefined processor that has content fetched (notably not NuGet at this time) has a new method that looks for interesting files and attaches their content to a hidden property on the document being processed
* The store configuration has a new chained store (content(<type>)) in the configuration. This watches for any document with the hidden property and stores all the objects attached there in a separate instance of the configured base store. 
* The default configuration of in Docker has been updated to have this new store setup
* Added several tests
* schema versions of all the ClearlyDefined tools updated

### DIFF
--- a/lib/baseHandler.js
+++ b/lib/baseHandler.js
@@ -47,11 +47,11 @@ class BaseHandler {
         (error, files) => {
           if (error) reject(error)
           if (files.length === 0) return resolve(null)
-          Object.defineProperty(document, '_fileContent', { value: [], enumerable: false })
+          Object.defineProperty(document, '_attachments', { value: [], enumerable: false })
           document.interestingFiles = files.map(file => {
-            const content = fs.readFileSync(path.join(location, file))
-            const token = BaseHandler.computeToken(content)
-            document._fileContent.push({ path: file, token, content })
+            const attachment = fs.readFileSync(path.join(location, file))
+            const token = BaseHandler.computeToken(attachment)
+            document._attachments.push({ path: file, token, attachment })
             return { path: file, token }
           })
           resolve()
--- a/lib/baseHandler.js
+++ b/lib/baseHandler.js
@@ -49,7 +49,7 @@ class BaseHandler {
           if (files.length === 0) return resolve(null)
           Object.defineProperty(document, '_attachments', { value: [], enumerable: false })
           document.interestingFiles = files.map(file => {
-            const attachment = fs.readFileSync(path.join(location, file))
+            const attachment = fs.readFileSync(path.join(location, file), 'utf8')
             const token = BaseHandler.computeToken(attachment)
             document._attachments.push({ path: file, token, attachment })
             return { path: file, token }
--- a/lib/baseHandler.js
+++ b/lib/baseHandler.js
@@ -6,6 +6,10 @@ const semver = require('semver')
 const EntitySpec = require('../lib/entitySpec')
 const extract = require('extract-zip')
 const decompress = require('decompress')
+const fs = require('fs')
+const path = require('path')
+const glob = require('glob')
+const shajs = require('sha.js')
 
 tmp.setGracefulCleanup()
 const tmpOptions = {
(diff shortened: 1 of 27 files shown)


providers/process/gemExtract.js
```diff
@@ -10,7 +10,7 @@ const sourceDiscovery = require('../../lib/sourceDiscovery')
 
 class GemExtract extends BaseHandler {
   get schemaVersion() {
-    return 1
+    return '1.1.0'
```
dabutvin: 👍

providers/process/npmExtract.js
```diff
@@ -28,10 +29,12 @@ class NpmExtract extends BaseHandler {
...
       const { document, spec } = super._process(request)
       this.addBasicToolLinks(request, spec)
-      const location = this._getManifestLocation(request.document.location)
-      const manifest = location ? JSON.parse(fs.readFileSync(location)) : null
+      const location = request.document.location
+      const manifestLocation = this._getManifestLocation(location)
+      const manifest = manifestLocation ? JSON.parse(fs.readFileSync(manifestLocation)) : null
       if (!manifest) console.log(`NPM without package.json: ${request.url}`)
-      await this._createDocument(request, manifest, request.document.registryData)
+      this._createDocument(request, manifest, request.document.registryData)
```
dabutvin: this method is still `async` - why wouldn't we want to await anymore?

lib/baseHandler.js
```diff
@@ -6,19 +6,60 @@ const semver = require('semver')
...
+    return shajs('sha256')
+      .update(content)
+      .digest('hex')
+  }
+
+  static addInterestingFiles(document, location) {
+    if (!location && location !== '') return null
+    return new Promise((resolve, reject) => {
+      glob(
+        buildGlob(['license', 'notice', 'notices']),
```
dabutvin: per the thread on discord add `COPYING` and `COPYING.LESSER` and all the files from scancode

```
    'copying',
    'copyright',
    'copyrights',
    'notice',
    'license',
    'licenses',
    'licence',
    'licences',
    'legal',
    'eula',
    'agreement',
    'copyleft',
    'licensing',
    'licencing',
    'patent',
    'patents',
```

https://github.com/nexB/scancode-toolkit/blob/develop/src/summarycode/classify.py#L133-L150

lib/baseHandler.js
```diff
@@ -6,19 +6,60 @@ const semver = require('semver')
...
   template: (process.platform === 'win32' ? 'c:/temp/' : '/tmp/') + 'cd-XXXXXX'
 }
 
+function buildGlob(roots) {
+  const parts = []
+  roots.forEach(root => {
+    parts.push(root)
+    parts.push(`${root}.{md,txt,html}`)
+  })
+  return `+(${parts.join('|')})`
```
dabutvin: do you only want to match files that are in the root?

`+(license|license.{md,txt,html}|notice|notice.{md,txt,html}|notices|notices.{md,txt,html})`

providers/process/nugetExtract.js
```diff
@@ -30,7 +30,7 @@ class NuGetExtract extends BaseHandler {
       const { spec } = super._process(request)
       this.addBasicToolLinks(request, spec)
       const manifest = await this._getManifest(request.document.location.manifest)
-      await this._createDocument(request, spec, manifest, request.document.registryData)
+      await this._createDocument(request, manifest, request.document.registryData)
```
dabutvin: it looks like nugetExtract is missing some updates.

 - update the schemaVersion to from `1` to `1.1.0`
 - add a call to `BaseHandler.addInterestingFiles`

The license files are typically dropped in the root of the nuget

for example, here is Newtonsoft.Json
image

lib/baseHandler.js
jeffmcaffer: Great list. I started to implement but it turns out that glob has some perf issues with that many combinations. Opened https://github.com/clearlydefined/crawler/issues/127 to follow up on this

jeffmcaffer:
OK. this is ready to merge IMHO. @dabutvin ? others?

providers/store/attachmentStore.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  upsert(document) {
+    const documentPromise = this.baseStore.upsert(document)
+    if (!document._attachments) return documentPromise
+    const attachmentPromises = document._attachments.map(entry => {
+      return this.baseStore.upsert({
+        _metadata: {
+          type: 'attachment',
+          url: `cd:/attachment/${entry.token}`,
```
dabutvin: is this supposed to be plural `/attachments/` ? 

https://github.com/clearlydefined/service/pull/193/files#diff-cb56bbc37b7fbad62f9e9ffd6c5f5877R39

dabutvin: oh i think not actually - it's for the coordinate not the actual route

https://github.com/clearlydefined/service/pull/193/files#diff-4bc830bb24cd56603b3dc4702dde1083R17

PR #128, opened 2018-08-15
Take 2 at capturing content

jeffmcaffer:
#125 proposes one approach to capturing interesting file content. structures the content as separate, globally deduplicated blobs. That is kind of nice but makes the system more complex as there are more stores to configure and keep in sync. It also means that the crawler would be outputting multiple files for one request. Possible but have to do trickery to get it done.

This PR is an alternative to that approach. Here we stash all the interesting content in the clearlydefined tool's output. This way the content is all together in one blob and the flow in the crawler is natural/normal.


jeffmcaffer:
After discussion and thought I think we should stick with the approach in #125. Better separation of concerns for the user and better caching and scale potential etc. Leaving this PR open until the other is merged in the event we change our minds.

jeffmcaffer:
closing in favor of #125

PR #131, opened 2018-08-21
add coordinates metadata to documents as they are processed

dabutvin:
We originally wanted to store this right in `_metadata/coordinates` for files and `_metadata/extra/coordinates` for blobs.
I ended up leaving the coordinates in `extra` all the time - after investigating, the complexity of differentiating where to store this data at this point did not seem worth it


dabutvin:
replaced with #132

PR #132, opened 2018-08-21
preserve name from registry in self link

dabutvin:


lib/baseHandler.js
```diff
@@ -136,22 +136,21 @@ class BaseHandler {
...
 
-  addSelfLink(request, urn = null) {
-    urn = urn || this.toSpec(request).toUrn()
-    request.linkResource('self', urn)
+  addSelfLink(request, preservedName = null) {
+    request.linkResource('self', this.getUrnFor(request, preservedName))
   }
 
   addBasicToolLinks(request, spec) {
-    request.linkResource('self', this.getUrnFor(request, spec))
```
dabutvin: call addSelfLink here

we need to preserve the siblings as well as the self link

lib/baseHandler.js
```diff
@@ -162,8 +162,7 @@ class BaseHandler {
     request.queue(name, spec.toUrl(), request.getNextPolicy(name))
   }
 
-  linkAndQueueTool(request, name, tool = name) {
-    const spec = this.toSpec(request)
+  linkAndQueueTool(request, spec, name, tool = name) {
```
jeffmcaffer: Just to confirm you looked at all callers to make sure they are not affected?

providers/process/pypiExtract.js
```diff
@@ -20,13 +20,15 @@ class PyPiExtract extends BaseHandler {
   }
 
   async handle(request) {
+    const spec = this.toSpec(request)
+    if (request.document.registryData.info.name) spec.name = request.document.registryData.info.name
```
jeffmcaffer: this could also be 

`spec.name = get(request, 'document.registryData.info.name') || spec.name`

Here and the similar pattern elsewhere.

lib/baseHandler.js
```diff
@@ -90,7 +90,7 @@ class BaseHandler {
 
   _process(request) {
     request.document._metadata.version = this.schemaVersion || 1
-    return { document: request.document, spec: this.toSpec(request) }
+    return { document: request.document }
```
jeffmcaffer: if we are just returning the document then perhaps just change to return the document rather than an object with the document in it. Alternatively, we can jst leave this doing toSpec for the normal case and allow people to recompute if needed.

providers/process/vsts.js
```diff
@@ -30,7 +30,7 @@ class VstsProcessor extends BaseHandler {
   }
 
   async handle(request) {
-    const { document } = super._process(request)
+    const document = super._process(request)
```
jeffmcaffer: _process still returns an object

lib/entitySpec.js
```diff
@@ -1,25 +1,6 @@
...
-const REVISION = 0x1
-const NONE = 0
-
-const toLowerCaseMap = {
-  github: NAMESPACE | NAME,
-  npmjs: NONE,
-  mavencentral: NONE,
-  mavencentralsource: NONE,
-  nuget: NAME
-}
```
jeffmcaffer: add pypi and rubygems (validate rubygems lowercasing NAME?)

Issue #133, opened 2018-08-22
maxBuffer exceeded for fossology / nomos scan 


dabutvin:
Queued an NPM component for harvesting and got this error output. Probably related to #130 

coordinate for repro: `npm/npmjs/-/grpc/1.12.3`

`Error fossology@cd:/npm/npmjs/-/grpc/1.12.3 stdout maxBuffer exceeded`

Full stack:

```
 Error: stdout maxBuffer exceeded
    at Socket.onChildStdout (child_process.js:328:14)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:246:13)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at Pipe.onread (net.js:597:20)
  cmd: 'cd /opt/fossology/src/nomos/agent && ./nomossa -ld /tmp/cd-JpP1Bv' 
```

```
curl -d '{"type":"npm", "url":"cd:/npm/npmjs/-/grpc/1.12.3"}' -H "Content-Type: application/json" -H "X-token: secret" -X POST http://crawlerhost/requests
```

PR #135, opened 2018-08-22
Fixed stdout maxBuffer exceeded issue

larainema:
Fix for issue #133 . The nomos output exceed the default stdout maxBuffer, not sure if it's a good fix, but didn't find any other good way to do it.


jeffmcaffer:
Thanks @larainema

PR #155, opened 2018-09-05
return null instead of empty array for default nuspec

dabutvin:
- [] does not meet the falsy check that follows in harvest()


dabutvin:
👍 👍

PR #158, opened 2018-09-07
update ghcrawler for improved deadletter messages

dabutvin:
see diff https://github.com/Microsoft/ghcrawler/compare/cfd8ffb...72cbb4e7


jeffmcaffer:
Seems like the something is broken here. the CI build is failing to clone the update-ghcrawler branch but it looks like maybe it is trying to get it out of the crawler repo rather than @dabutvin 's fork. I tried requeuing the build but that did not help.

PR #173, opened 2018-10-08
make sure registryData is defined before getting nupkg

dabutvin:


providers/fetch/nugetFetch.js
```diff
@@ -30,7 +30,7 @@ class NuGetFetch extends BaseHandler {
       throw new Error('NuGet package could not be detected probably due to non-existent revision or name.')
     const dir = this._createTempDir(request)
     const location = await this._persistMetadata(dir, manifest, nuspec)
-    location.nupkg = await this._getNupkg(dir, registryData.packageContent)
+    location.nupkg = registryData ? await this._getNupkg(dir, registryData.packageContent) : null
```
jeffmcaffer: Just as a point of interest, is the extractor written to handle null for `location.nupkg`

dabutvin: yep. should be good
https://github.com/clearlydefined/crawler/blob/master/lib/baseHandler.js#L43

PR #183, opened 2018-10-16
fork decompress-unzip to turn off validation #178

### LINKED ISSUE #178: compressed/uncompressed size mismatch unzipping some nugets
The errors look like:

```
compressed/uncompressed size mismatch for stored file: 2 != 0
```

This is coming from an optional validation in `yauzl` see https://github.com/thejoshwolfe/yauzl/issues/53

They have an option to disable this validation, which we could use since we are not in control of the format of these zips.

The tricky part is we are using `yauzl` through https://github.com/kevva/decompress-unzip/blob/master/package.json which is not up to date with the version that exposes it as an option.

The option would go [here](https://github.com/kevva/decompress-unzip/blob/master/index.js#L85) next to `lazyEntries` 

it would be

```
validateEntrySizes: false
```

dabutvin:
fix #178

### DIFF
--- a/lib/baseHandler.js
+++ b/lib/baseHandler.js
@@ -6,6 +6,8 @@ const semver = require('semver')
 const EntitySpec = require('../lib/entitySpec')
 const extract = require('extract-zip')
 const decompress = require('decompress')
+const decompressTargz = require('decompress-targz')
+const decompressUnzip = require('decompress-unzip')
 const fs = require('fs')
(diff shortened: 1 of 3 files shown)


dabutvin:
have a PR going upstream: https://github.com/kevva/decompress-unzip/pull/21

dabutvin:
@jeffmcaffer want to look this over before we roll forward?

PR #192, opened 2018-11-06
add quotes around test path

dabutvin:
wasn't finding all the tests on `npm test` without this


providers/process/sourceExtract.js
```diff
@@ -32,7 +32,7 @@ class SourceExtract extends BaseHandler {
     await BaseHandler.addInterestingFiles(request.document, location)
     const clearlyFile = path.join(location, 'clearly.yaml')
     if (!fs.existsSync(clearlyFile)) return
-    const content = await promisfy(fs.readFileSync)(clearlyFile)
+    const content = await promisify(fs.readFileSync)(clearlyFile)
```
dabutvin: this was a bad one that lint caught

PR #193, opened 2018-11-07
add more license filenames

dabutvin:
these seem to be common filenames that keep popping up that we are missing right now


dabutvin:
If we want to do this, we will need to update the service IsLicenseFile() check here: https://github.com/clearlydefined/service/blob/3eeade82e03b9bf955ba40897434d1579fdd1e10/lib/utils.js#L169

dabutvin:
I am going to run some benchmarks for globbing these files to validate we won't have any perf issues by extending this list

dabutvin:
Here is the benchmark results

25 runs. For each file we glob it extension-less and also with 3 other extensions - so 4X patterns , take the average of all 25 runs. Node glob scaled exponentially and then crashed. fast-glob stayed flat

image

# node-glob

- [https://www.npmjs.com/package/glob](https://www.npmjs.com/package/glob)
- [https://github.com/isaacs/node-glob](https://github.com/isaacs/node-glob)

image


# fast-glob

- [https://www.npmjs.com/package/fast-glob](https://www.npmjs.com/package/fast-glob)
- [https://github.com/mrmlnc/fast-glob](https://github.com/mrmlnc/fast-glob)

image

lib/baseHandler.js
```diff
@@ -42,31 +42,27 @@ class BaseHandler {
...
-              return { path: relativePath, token, license }
-            })
-          ).then(interestingFiles => {
-            document.interestingFiles = interestingFiles
-            resolve()
-          })
-        }
-      )
+    if (!location && location !== '') return Promise.resolve(null)
+    return glob(buildGlob(['license', 'notice', 'notices', 'license-mit', 'license-apache', 'unlicense']), {
```
jeffmcaffer: Why not async/await pattern?

PR #196, opened 2018-11-08
release to prod

dabutvin:


providers/process/crateExtract.js
```diff
@@ -0,0 +1,57 @@
...
+  async handle(request) {
+    if (this.isProcessing(request)) {
+      const { spec } = super._process(request)
+      this.addBasicToolLinks(request, spec)
+      const location = request.document.location
+      const manifest = request.document.manifest
+      await this._createDocument(request, manifest, request.document.registryData)
+      await BaseHandler.addInterestingFiles(request.document, location)
+    }
+    this.linkAndQueueTool(request, 'scancode')
```
dabutvin: this is not being queued quite right.

the scancode output is losing a tool version
image

dabutvin: this is actually a bug in the initialization of the scancode processor

image

we might be able to hang the processor until that's complete, or better yet, set it when the container installs scancode maybe?

opened #197 to track

PR #204, opened 2018-11-12
update 404s to markskip when returning

dabutvin:


dabutvin:
verified each of these are behaving appropriately

@jeffmcaffer fyi

PR #208, opened 2018-11-22
merge the pr and dev ci build definitions

jeffmcaffer:


jeffmcaffer:
abandoning this approach so we have individual build defs in the UI for tracking etc.

PR #209, opened 2018-11-27
console.log by default

dabutvin:
- add LOGGING_STDOUT_OFF to disable


jeffmcaffer:
I proposed some other changes in https://github.com/clearlydefined/crawler/tree/dabutvin-stdout. Mostly just normalizing the ENV var naming and access techniques. @dabutvin see what you think and include here as desired.

PR #216, opened 2018-12-02
Add support for FOSSology's Monk and Copyright tools 

### LINKED ISSUE #150: Integrate Monk
Once https://github.com/fossology/fossology/pull/1162 is merged, we can integrate Monk in the same way we did Nomos.

jeffmcaffer:
fixes #150 

Monk is very much like the copyright agent so do the same thing.

Some remaining points to consider
* need to generate and store the license database. likely in the docker build. Not really sure where all the data comes from.
* proposed implementation here runs monk once for each file. That causes the license DB to get loaded once per file. Likely a lot of overhead. Monk can take a list of files. Little unclear what the output looks like (JSON one entry per line with no info on about which file?). May be worth looking at running with more than one file at a time

Note that this depends on #228

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,8 @@ RUN gem install licensee -v 9.10.1 --no-rdoc --no-ri
 WORKDIR /opt
 RUN git clone https://github.com/fossology/fossology.git
 
+# See https://github.com/fossology/fossology/blob/faaaeedb9d08f00def00f9b8a68a5cffc5eaa657/utils/fo-installdeps#L103-L105
+# Additional libjsoncpp-dev https://github.com/fossology/fossology/blob/261d1a3e663b5fd20652a05b2d6360f4b31a17cb/src/copyright/mod_deps#L79-L80
 RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests \
   libmxml-dev curl libxml2-dev libcunit1-dev libjsoncpp-dev \
   build-essential libtext-template-perl subversion rpm librpm-dev libmagic-dev libglib2.0 libboost-regex-dev libboost-program-options-dev
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,14 +31,21 @@ RUN gem install licensee -v 9.10.1 --no-rdoc --no-ri
 WORKDIR /opt
 RUN git clone https://github.com/fossology/fossology.git
 
+RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests \
+  libmxml-dev curl libxml2-dev libcunit1-dev libjsoncpp-dev \
+  build-essential libtext-template-perl subversion rpm librpm-dev libmagic-dev libglib2.0 libboost-regex-dev libboost-program-options-dev
+
 WORKDIR /opt/fossology/src/nomos/agent
 RUN make -f Makefile.sa
+RUN echo $(./nomossa -V)
 
+# NOTE: must build copyright before Monk to cause libfossology to be built
 WORKDIR /opt/fossology/src/copyright/agent
 RUN make
 
 WORKDIR /opt/fossology/src/monk/agent
 RUN make
+RUN echo $(./monk -V)
 
 ENV FOSSOLOGY_HOME=/opt/fossology/src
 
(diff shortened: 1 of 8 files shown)


jeffmcaffer:
@maxhbr, any thoughts on where/how we can get a license database to pass in to `monk -k`. In theory we could run `monk -s` first but that is going to want a database etc. We could also get the Monk image and run it in the docker build, run `-s` and then grab the output file. Would be a little concerned about version mismatch...

maxhbr:
One could try to add a make target, which generates the dump. E.g. by extending the monk code one could write something, which parses an input file to generate the dump (instead of depending on the database).

But parsing the FOSSology input file https://raw.githubusercontent.com/fossology/fossology/master/install/db/licenseref.sql might be hard. To simplify the problem one could refactor this input to something easier to parse. E.g. `JSON` or `CSV` and generate the `sql` file from that.

jeffmcaffer:
Thanks @maxhbr. How does `licenseref.sql` get created? It should be reasonably easy to parse but I feel like people are not hand editing that file to add a new license. Looking through the code I don't see any place where it is created. Just is assumed to be there.

larainema:
@jeffmcaffer `licenseref.sql` is used to create the license_ref table when initial install the FOSSology, it can be generated by the script https://github.com/fossology/fossology/blob/master/install/db/export_license_ref.php from the database after user edit the license_ref table.

jeffmcaffer:
Thanks @larainema . So the user experience for adding a license is editing a table in the database? I guess there is a UI for this? I'm guessing that the user adds a license and then runs this script and checks the resulting file into the Git repo for future runs. On first startup it looks like this file is also used to initialize the database. So it's sort of circular (not in a bad way).

Anyway, perhaps something like https://dba.stackexchange.com/questions/90482/export-postgres-table-as-json could be used to dump in json format. We probably don't need a whole lot of re-engineering here, just a decent way of getting a file that can be used with `monk -k`. Preferably we would not have to have Postgres or whole FOSSology instance just to get this file. If the json dump (whatever) could be checked in beside the current `license_ref.sql` file, that would work.

providers/process/fossology.js
```diff
@@ -89,33 +142,65 @@ class FossologyProcessor extends BaseHandler {
...
+  }
+
   async _createDocument(request) {
-    //const files = await getFiles(request.document.location)
+    const files = await getFiles(request.document.location)
     const nomosOutput = await this._runNomos(request)
-    //const copyrightOutput = await this._runCopyright(request, files)
-    //const monkOutput = await this._runMonk(request, files)
+    const copyrightOutput = await this._runCopyright(request, files)
+    const monkOutput = await this._runMonk(request, files)
```
dabutvin: if either of these 3 tools throw do we want to produce a partial document?
Or is the plan to be all or none for fossology?

providers/process/fossology.js
```diff
@@ -49,25 +41,86 @@ class FossologyProcessor extends BaseHandler {
...
+          }
+          resolve(stdout)
+        }
+      )
+    })
+  }
+
+  async _runMonk(request, files) {
+    // TODO can't actually run Monk until the license database is factored out
+    return null
```
dabutvin: are we going to block on this, or should we open an issue to follow up on it?

providers/process/fossology.js
```diff
@@ -49,25 +41,86 @@ class FossologyProcessor extends BaseHandler {
...
-        }
-        let buffer = new Buffer(stdout)
-        buffer = bufferReplace(buffer, request.document.location + '/', '')
-        const nomosOutput = {
-          version: this.nomosVersion,
-          parameters: parameters,
-          output: {
+      const parameters = [].join(' ')
+      exec(
+        `cd ${this.options.installDir}/nomos/agent && ./nomossa -ld ${request.document.location} ${parameters}`,
```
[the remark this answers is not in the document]
dabutvin: this makes sense with the pattern 👍

providers/process/fossology.js
dabutvin: all or none ftw!

providers/process/fossology.js
dabutvin: see #240

Dockerfile
```diff
@@ -31,10 +31,25 @@ RUN gem install licensee -v 9.10.1 --no-rdoc --no-ri
 WORKDIR /opt
 RUN git clone https://github.com/fossology/fossology.git
```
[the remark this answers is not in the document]
dabutvin: #241

Issue #222, opened 2018-12-10
 Check the DQ count when pulling from the queue


dabutvin:
Set a limit of retries (5?) 

We are seeing 20+ DQs on a single message.
When it passes the threshold, mark it dead and move on so we do not hold upthe queue and waste cpu

PR #223, opened 2018-12-11
set maxDequeueCount to 5

dabutvin:
see https://github.com/Microsoft/ghcrawler/pull/161 for new config


dabutvin:
fixes https://github.com/clearlydefined/crawler/issues/222

PR #232, opened 2018-12-20
CocoaPods integration

Frugghi:
I couldn't find a way to implement `top` for CocoaPods so I've not implemented it. They only have an [API](https://github.com/CocoaPods/metrics.cocoapods.org) for getting metrics about a specific pod.

### DIFF
--- a/providers/fetch/podFetch.js
+++ b/providers/fetch/podFetch.js
@@ -1,11 +1,19 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-const { clone } = require('lodash')
+const { clone, get } = require('lodash')
 const BaseHandler = require('../../lib/baseHandler')
 const request = require('request-promise-native')
 const fs = require('fs')
 const path = require('path')
+const crypto = require('crypto')
+const { exec } = require('child_process')
+const requestRetry = require('requestretry').defaults({ maxAttempts: 3, fullResponse: true })
+
+const services = {
+  trunk: 'https://trunk.cocoapods.org/api/v1',
+  specs: 'https://raw.githubusercontent.com/CocoaPods/Specs/master'
+}
 
 class PodFetch extends BaseHandler {
   canHandle(request) {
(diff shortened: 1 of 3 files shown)


providers/fetch/podFetch.js
```diff
@@ -15,59 +23,149 @@ class PodFetch extends BaseHandler {
...
 
   async _getRegistryData(spec) {
     let registryData
     try {
-      // registryData = await request({
-      //   url: `https://crates.io/api/v1/crates/${spec.name}`,
-      //   json: true
-      // })
+      registryData = await request({
+        url: `${services.specs}/Specs/${this._masterRepoPathFragment(spec, [1, 1, 1])}/${spec.name}.podspec.json`,
```
dabutvin: this endpoint is going to be ratelimited at 60 requests per hour if we do it unauthenticated.

see `cdConfig.js` for how we can pass a githubToken in here to bump that limit up to 5000

providers/fetch/podFetch.js
```diff
@@ -15,59 +23,149 @@ class PodFetch extends BaseHandler {
...
+
+  async _getPackage(dir, podspec) {
+    const httpSource = get(podspec, 'source.http')
+    const gitSource = get(podspec, 'source.git')
+    if (httpSource) {
+      return await this._getSourceArchive(dir, httpSource, podspec)
+    } else if (gitSource) {
+      return await this._getGitRepo(dir, gitSource, podspec)
+    } else {
+      // Unsupported source (e.g. SVN and Mercurial)
```
dabutvin: i wonder how common this is - seems like most are using `git`

providers/fetch/podFetch.js
```diff
@@ -15,59 +23,149 @@ class PodFetch extends BaseHandler {
...
-      // version: registryData.versions.find(x => x.num === version)
+
+    return registryData
+  }
+
+  async _getPackage(dir, podspec) {
+    const httpSource = get(podspec, 'source.http')
+    const gitSource = get(podspec, 'source.git')
+    if (httpSource) {
+      return await this._getSourceArchive(dir, httpSource, podspec)
```
dabutvin: do you have an example pod that uses this unzip?

providers/fetch/podFetch.js
```diff
@@ -15,59 +23,149 @@ class PodFetch extends BaseHandler {
...
+      )
+    })
+  }
+
+  async _getGitRepo(dir, repo, podspec) {
+    const rev = get(podspec, 'source.commit') || get(podspec, 'source.tag')
+
+    if (!rev) {
+        // CocoaPods support source.branch too
+        return null
```
dabutvin: i think this makes sense and is the right call, we can't harvest a branch that could change underneath us

providers/fetch/podFetch.js
Frugghi: `Fabric` and `Crashlytics`

providers/fetch/podFetch.js
Frugghi: I've never seen a pod not using git or http sources, I'm sure they aren't very popular!

providers/fetch/podFetch.js
Frugghi: That's the reason why I didn't include it. I think most of the libraries specify a tag or a commit, the branch option shouldn't be very popular.

Issue #182, opened 2018-10-13
Many temp 503s from Maven search


dabutvin:
503 Service Temporarily Unavailable

we are hitting urls like 

 - https://search.maven.org/solrsearch/select?q=g:"commons-logging"+AND+a:"commons-logging"+AND+v:"1.1"&rows=1&wt=json

We are probably overloading the server, we could add some kind of throttle or retry backoff policy

dabutvin:
this is still an issue for getting poms using something like https://search.maven.org/remotecontent?filepath=com/weicoder/redis/3.0.0/redis-3.0.0.pom

dabutvin:
we could try using a different url like: https://repo1.maven.org/maven2/com/weicoder/redis/3.0.0/redis-3.0.0.pom

this may be more direct

nellshamrell:
At this time I cannot reproduce this. If anyone sees this in the future, please feel free to re-open this and we will investigate further.

PR #234, opened 2019-01-01
refactor maven support

jeffmcaffer:
Refactor the maven support to be simpler and more like the other providers. Also add better testing.


providers/fetch/mavencentralFetch.js
```diff
@@ -0,0 +1,120 @@
...
+    if (!registryData) return request.markSkip('Missing  ')
+    spec.revision = spec.revision ? registryData.v : registryData.latestVersion
+    // rewrite the request URL as it is used throughout the system to derive locations and urns etc.
+    request.url = spec.toUrl()
+    super.handle(request)
+    const poms = await this._getPoms(spec)
+    if (!poms.length) return request.markSkip('Missing  ')
+    const artifact = this.createTempFile(request)
+    const code = await this._getArtifact(spec, artifact.name)
+    if (code === 404) return request.markSkip('Missing  ')
```
dabutvin: Do you think we should we make a method on baseHandler so we can call `this.skip()`

```js
skip(request) {
  return request.markSkip('Missing  ')
}
```

providers/fetch/mavencentralFetch.js
```diff
@@ -0,0 +1,120 @@
...
+    const dir = this.createTempDir(request)
+    // Warning: may not clean files up on Windows due to a bug. Switch back to unzip once https://github.com/maxogden/extract-zip/issues/65 is resolved
```
dabutvin: This is fixed in #176 
We are having good success with decompress in it's current configuration
 do you think it is still worth trying to switch back to extract-zip?

(They both use yauzl under the hood)

providers/fetch/mavencentralFetch.js
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  // query maven to get the latest version if we don't already have that.
+  // Example: https://search.maven.org/solrsearch/select?q=g:%22org.eclipse%22+AND+a:%22swt%22+AND+v:%223.3.0-v3346%22&rows=1&wt=json
+  async _getRegistryData(spec) {
+    const versionClause = spec.revision ? `+AND+v:"${spec.revision}"` : ''
+    const url = `https://search.maven.org/solrsearch/select?q=g:"${spec.namespace}"+AND+a:"${
+      spec.name
+    }"${versionClause}&rows=1&wt=json`
+    return await requestPromise({ url, json: true })
```
dabutvin: I wonder if this is a good opportunity to take on #182 ?

My hunch is that some kind of throttle/bottleneck that tracks between requests per instance is all we need.

providers/fetch/mavencentralFetch.js
```diff
@@ -0,0 +1,120 @@
...
+          if (response.statusCode !== 200) reject(new Error(`${response.statusCode} ${response.statusMessage}`))
+        })
+        .pipe(fs.createWriteStream(destination).on('finish', () => resolve(null)))
+    })
+  }
+
+  async _getPoms(spec, result = []) {
+    const pom = await this._getPom(spec)
+    const parentSpec = this._buildParentSpec(pom)
+    if (parentSpec) await this._getPoms(parentSpec, result)
```
dabutvin: this is cool - is the plan for the summarizer in the service to read through these to get declared license from parents?

are you worried about this getting circular at all?

providers/process/mavenExtract.js
```diff
@@ -107,8 +66,10 @@ class MavenExtract extends AbstractClearlyDefinedProcessor {
       githubToken: this.options.githubToken
     })
     if (githubSource) return githubSource
-    // didn't find any source so make up a sources url to try if the registry thinks there is source
-    if (!registryData.ec || !registryData.ec.includes(mavenCentral.sourceExtension)) return null
+    // didn't find any source in GitHub so make up a sources url to try if the registry thinks there is source
+    // TODO could check `registryData.ec.includes('-sources.jar')` but there seemed to be examples where
+    // the registry did not think it had the sources jar but it did. So Just make it up here and we'll try.
+    // Need to confirm the expectations here.
```
dabutvin: this list may be interesting, coming from the search endpoint

```
"-javadoc.jar",
"-sources.jar",
"-test-sources.jar",
".jar",
"-tests.jar",
"-api.jar",
"-src.zip",
".pom",
"-src.tar.gz",
"-adapters.jar",
"-bin.zip",
"-bin.tar.gz"
```

providers/fetch/mavencentralFetch.js
dabutvin: better yet, let's only call this endpoint if there is no revision specified.
We can get the casedSpec from the POM

providers/process/mavenExtract.js
```diff
@@ -46,52 +46,11 @@ class MavenExtract extends AbstractClearlyDefinedProcessor {
...
-    const poms = manifest.poms.slice(0)
-    poms.unshift(pom)
-    return {
-      summary: summary,
-      poms: poms
-    }
+  _mergePoms(poms) {
+    if (!poms) return null
+    return [...poms].reverse().reduce((result, pom) => {
+      return { ...result, ...pom.project }
```
dabutvin: I'm seeing an issue with this where a child pom can have a property value of `parent` that seems to point to a parent pom, but this will overwrite the parent as this is currently coded

```json
poms[0]
{
    "project": {
      "modelVersion": ["4.0.0"],
      "groupId": ["com.weicoder"],
      "artifactId": ["parent"],
      "version": ["3.0.0"],
      "packaging": ["pom"],
      "name": ["parent"],
      "description": ["java tools"],
      "url": ["https://github.com/wdcode/weicoder"],
      "properties": [
        {
          "project.build.sourceEncoding": ["UTF-8"],
          "project.reporting.outputEncoding": ["UTF-8"],
          "parent.version": ["3.0.0"]
        }
      ],
```

```json
poms[1]
{
    "project": {
      "parent": [
        {
          "groupId": ["com.weicoder"],
          "artifactId": ["parent"],
          "version": ["3.0.0"],
          "relativePath": ["../parent/pom.xml"]
        }
      ],
      "modelVersion": ["4.0.0"],
      "artifactId": ["redis"],
      "packaging": ["jar"],
      "name": ["redis"],
      "description": ["java redis"],
      "url": ["https://github.com/wdcode/weicoder"]
```

this gives us a 'merged' pom of

```json
{
  "parent": [
    {
      "groupId": ["com.weicoder"],
      "artifactId": ["parent"],
      "version": ["3.0.0"],
      "relativePath": ["../parent/pom.xml"]
    }
  ],
  "modelVersion": ["4.0.0"],
  "artifactId": ["parent"],
  "packaging": ["pom"],
  "name": ["parent"],
  "description": ["java tools"],
  "url": ["https://github.com/wdcode/weicoder"],
  "groupId": ["com.weicoder"],
  "version": ["3.0.0"],
  "properties": [
    {
      "project.build.sourceEncoding": ["UTF-8"],
      "project.reporting.outputEncoding": ["UTF-8"],
      "parent.version": ["3.0.0"]
    }
  ],
```

dabutvin: they just needed to be un-reversed :) i took care of it

PR #238, opened 2019-01-08
delete .git folder in gitcloner when we are done

dabutvin:


providers/fetch/gitCloner.js
```diff
@@ -59,6 +59,14 @@ class GitCloner extends BaseHandler {
     // ...\nsize-pack: 3\n... (in KB)
     return Number(gitCountObjectsResult.match('size-pack: (.*)\n')[1])
   }
+
+  _deleteGitDatabase(dirName, specName) {
+    return new Promise((resolve, reject) => {
```
jeffmcaffer: Could also use `promisify`

PR #264, opened 2019-02-22
expand filenames to harvest (fixes #127)

### LINKED ISSUE #127: Enhance the list of interesting files
The crawler can capture the content of *interesting* files (e.g., LICENSE). Currently it captures just a few. ScanCode looks at bigger list as interesting. 
```javascript
      const filenames = [
        'agreement',
        'copying',
        'copyleft',
        'copyright',
        'copyrights',
        'eula',
        'license',
        'licenses',
        'licensing',
        'licence',
        'licences',
        'licencing',
        'legal',
        'notice',
        'notices',
        'patent',
        'patents'
      ]
```
Typically a given component has only one or two of these files but we need to look for all just in case.

Note that that current implementation uses `glob` to look at the file system. It appears that this is quite expensive if you construct a glob expression that can handle all the variations between these files and the possible extensions. We need to either change the way the glob expression is structured or use a different approach. For example, we can leverage the following to do something really simple

1. we are only looking at the root folder of the component
2. The list of files above is large but not that large. Could just generate all the combinations and do a simple dir list, toLowerCase and then intersect.

Note also that we should include `.rst` as another possible extension

dabutvin:
I figured we should just take all file extensions for attaching and be a little more strict when assigning nature's

### DIFF
--- a/providers/process/abstractClearlyDefinedProcessor.js
+++ b/providers/process/abstractClearlyDefinedProcessor.js
@@ -8,6 +8,30 @@ const path = require('path')
 const { pick, merge } = require('lodash')
 const du = require('du')
 const { trimParents } = require('../../lib/utils')
+const interestingFileNames = new Set([
+  'AGREEMENT',
+  'CONTRIBUTORS',
+  'COPYING',
+  'COPYLEFT',
+  'COPYRIGHT',
+  'COPYRIGHTS',
+  'EULA',
+  'LEGAL',
+  'LICENCE',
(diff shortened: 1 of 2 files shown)


geneh:
This PR should further be tested and merged.

nellshamrell:
Due to the age of this pull request, I'm going to go ahead and close it. However, @dabutvin, if you wish to pursue this further please feel free to re-open it and we can discuss!

PR #266, opened 2019-02-28
Kubernetes: Add resource requests and limits and a horizonal autoscaler.

rspier:
This should allow resources to fluctuate up and down as the crawler backlog
varies based on aggregate CPU usage.

Also commit some old documentation notes.


dabutvin:
👍

PR #270, opened 2019-03-04
use smartgit to discoverFromGitHubTags (fixes #269)

### LINKED ISSUE #269: use smartgit protocol for source discovery from github
we are using api.github.com to associate tags to commit hashes.
We can use the same protocol cloning uses instead. Since these are all public repos, we won't need to use any tokens

for example:

today we are using 
https://api.github.com/repos/isaacs/node-glob/tags

we can switch to
https://github.com/isaacs/node-glob.git/info/refs?service=git-upload-pack

at https://github.com/clearlydefined/crawler/blob/master/lib/sourceDiscovery.js

dabutvin:
the url internally is
https://github.com/{owner}/{repo}.git/info/refs?service=git-upload-pack

- uses https://www.npmjs.com/package/geit to fetch refs instead gh apis
- will not rely on gh tokens this way

### DIFF
--- a/lib/sourceDiscovery.js
+++ b/lib/sourceDiscovery.js
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-const _ = require('lodash')
+const { uniq, uniqWith } = require('lodash')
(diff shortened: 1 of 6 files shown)


dabutvin:
I verified this with a large set of existing data.

Issue #252, opened 2019-01-25
[Fossology] protect file arguments for spaces and special characters


dabutvin:
Getting this error. Need to properly quote/escape the paths passed into exec

```
Command failed: cd /opt/fossology/src/copyright/agent &&
./copyright --files /tmp/cd-w7XBQt/package/ios/Vendor/SDWebImage/Examples/SDWebImage TV Demo/Assets.xcassets/App Icon & Top Shelf Image.brandassets/App Icon - Large.imagestack/Middle.imagestacklayer/Content.imageset/Contents.json -J

/bin/sh: 1: Top: not found
```

dabutvin:
we should go through every exec and protect. Treating the arguments as user input.

dabutvin:
im working on this one right now

dabutvin:
this is done

PR #287, opened 2019-04-11
move exec inputs to parameter lists from string manipulations

dabutvin:
we handle parameters and arguments as arrays of individual values instead of executing a string. This is to prevent malicious command injection, and also to help handle file/folder names that are not compatible with the shell.


dabutvin:
fixes #252

Issue #292, opened 2019-04-24
Some attachments and scan results exceed 64MB


dabutvin:
```
RangeError: createBlockBlobFromText requires the size of text to be less than 64MB. Please use createBlockBlobFromLocalFile or createBlockBlobFromStream to upload large blobs.

    at Object.BlobService.createBlockBlobFromText (/opt/service/node_modules/azure-storage/lib/services/blob/blobservice.js:3109:15)
    at AzureStorageDocStore.upsert (/opt/service/node_modules/ghcrawler/providers/storage/storageDocStore.js:50:18)
    at AttachmentStore.upsert (/opt/service/providers/store/attachmentStore.js:17:44)
```

see https://github.com/clearlydefined/crawler/blob/ffdc7e139bee61b30d308a3c5a53a1a0ba6e0222/providers/store/attachmentStore.js#L17

This would need to be updated in ghcrawler if we want to support huge attachments.
We could also do validation for attachmentSize if we do not want to add support for huge attachments

geneh:
We should support it.

PR #312, opened 2019-06-13
Support blobs over 64 MB

geneh:
- Point to the latest commit in ghcrawler lib
- Remove base-64 package to resolve InvalidCharacterError (The string to be encoded contains characters outside of the Latin1 range.)


geneh:
Resolves https://github.com/clearlydefined/crawler/issues/292

PR #383, opened 2020-03-12
Only shallow clone repos to save on time we spend cloning git

brphelps:


jeffmendoza:
Not sure about this until we have a better way to monitor crawler errors. Right now it is very manual

PR #424, opened 2021-04-06
Initial changes to support mave.google.com

pnibakuze-ms:
Work in progress:

Added mavengoogle as new provider to avoid making lots of changes on Maven Central fetch module
Created a new module mavenGoogleFetch for harvesting data from maven.google.com
Because there are no API, I had to use Chrome dev tool to analyze how requests are being sent.
All packages follow the same structure with following example format:
https://maven.google.com/web/index.html#android.arch.navigation:navigation-common:1.0.0
To extract actual source file, the URL is format as follows:
https://dl.google.com/android/maven2/android/arch/navigation/navigation-common/1.0.0/navigation-common-1.0.0.aar
For testing, send the following request:
http://localhost:3000/definitions/maven/mavengoogle/android.arch.navigation/navigation-common/1.0.0
Or
curl http://localhost:4000/definitions/maven/mavengoogle/android.arch.navigation/navigation-common/1.0.0


nellshamrell (commented):
This looks FANTASTIC and I really appreciate the thorough tests!

nellshamrell (approved):
This is GREAT! Thank you again! Approving the changes, but holding off on merging until we are able to talk to someone at Google (which is in process behind the scenes).

PR #458, opened 2022-01-25
add patch-package to be installed outside of dev environments as well

nellshamrell:
Signed-off-by: Nell Shamrell <nells@microsoft.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,13 +52,13 @@
     "lodash": "^4.17.21",
     "luxon": "^2.1.1",
     "lzma-native": "^8.0.1",
-    "luxon": "^2.1.1",
     "memory-cache": "^0.2.0",
     "mkdirp": "^0.5.1",
     "morgan": "^1.9.0",
     "node-uuid": "^1.4.8",
     "painless-config": "^0.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -52,13 +52,13 @@
     "lodash": "^4.17.21",
     "luxon": "^2.1.1",
     "lzma-native": "^8.0.1",
-    "luxon": "^2.1.1",
```
nellshamrell, line 55: Looks like luxon was in the package.json file twice. Not sure how it got that way, but this fixes it!

package.json
```diff
@@ -52,13 +52,13 @@
...
     "memory-cache": "^0.2.0",
     "mkdirp": "^0.5.1",
     "morgan": "^1.9.0",
     "node-uuid": "^1.4.8",
     "painless-config": "^0.1.2",
     "parse-github-url": "^1.0.2",
+    "patch-package": "^6.2.2",
```
nellshamrell, line 61: Moved this from the dev dependencies to the "regular" dependencies - this appears to be needed in our Azure DevOps builds after upgrading to Node 16.

disulliv (approved):

PR #460, opened 2022-02-04
Additional enhancements for the upgraded ScanCode

nellshamrell:
Adding some additional enhancements after upgrading Scancode.

Based on comments from @pombredanne in #459

### DIFF
--- a/DevDockerfile
+++ b/DevDockerfile
@@ -8,7 +8,7 @@ ARG BUILD_NUMBER=0
 ENV CRAWLER_BUILD_NUMBER=$BUILD_NUMBER
 
 # Ruby and Python Dependencies
-RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests curl bzip2 build-essential libssl-dev libreadline-dev zlib1g-dev cmake python3 python3-dev python3-pip && \
+RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests curl bzip2 build-essential libssl-dev libreadline-dev zlib1g-dev cmake python3 python3-dev python3-pip xz-utils libxml2-dev libxslt1-dev libpopt0 && \
   rm -rf /var/lib/apt/lists/* && \
   curl -L https://github.com/rbenv/ruby-build/archive/v20180822.tar.gz | tar -zxvf - -C /tmp/ && \
   cd /tmp/ruby-build-* && ./install.sh && cd / && \
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -62,7 +62,7 @@ module.exports = {
     pod: { githubToken },
     pypi: { githubToken },
     scancode: {
-      installDir: config.get('SCANCODE_HOME') || 'C:\\installs\\scancode-toolkit-3.0.2',
+      installDir: config.get('SCANCODE_HOME'),
       options: [
         '--copyright',
         '--license',
@@ -70,6 +70,7 @@ module.exports = {
         '--license-text',
         '--is-license-text',
         '--package',
+        '--license-text-diagnostics',
         '--strip-root',
         '--email',
         '--url',
--- a/providers/process/scancode.js
+++ b/providers/process/scancode.js
@@ -108,20 +108,13 @@ class ScanCodeProcessor extends AbstractProcessor {
 
   _detectVersion() {
     if (this._versionPromise) return this._versionPromise
-
-    // Create trivial file to run Scancode on so we can get the version
-    // of it in a safer way than parsing stdout
-
-    fs.writeFileSync('NOTICE', 'Sample text')
-
-    this._versionPromise = execFile(`${this.options.installDir}/scancode`, ['--json-pp', '-', 'NOTICE'])
+    this._versionPromise = execFile(`${this.options.installDir}/scancode`, ['--version'])
       .then(result => {
         this.logger.info('Detecting ScanCode version')
 
         const raw_output = result.stdout
-        const scanCodeOutputJson = JSON.parse(raw_output)
-        this._toolVersion = scanCodeOutputJson.headers[0].tool_version
-
+        const scancode_line = raw_output.match(/ScanCode version .*\n/)[0]
+        this._toolVersion = scancode_line.replace('ScanCode version ', '').trim()
         this._schemaVersion = this.aggregateVersions(
           [this._schemaVersion, this.toolVersion, this.configVersion],
           'Invalid ScanCode version'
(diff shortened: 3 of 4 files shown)


config/cdConfig.js
```diff
@@ -62,14 +62,15 @@ module.exports = {
     pod: { githubToken },
     pypi: { githubToken },
     scancode: {
-      installDir: config.get('SCANCODE_HOME') || 'C:\\installs\\scancode-toolkit-3.0.2',
+      installDir: config.get('SCANCODE_HOME'),
```
nellshamrell, line 65: I removed the default installDir for ScanCode - this is going to vary so much based on what OS someone is developing with and will change every time ScanCode is updated.

providers/process/scancode.js
```diff
@@ -108,13 +108,27 @@ class ScanCodeProcessor extends AbstractProcessor {
 
   _detectVersion() {
     if (this._versionPromise) return this._versionPromise
-    this._versionPromise = execFile(`${this.options.installDir}/scancode`, ['--version'])
+
+    // Create trivial file to run Scancode on so we can get the version
+    // of it in a safer way than parsing stdout
+    let fileCreationErrors = ''
+
+    child_process.exec('touch NOTICE', (error) => {
```
nellshamrell, line 116: The ScanCode version is only detected when the application is initialized, so this file creation will only happen when the container is first started. This seemed to be the best way to have something to run ScanCode on to get the tool version, rather than trying to parse the output of the `scancode --version` command (which is very brittle).

providers/process/scancode.js
```diff
@@ -108,13 +108,27 @@ class ScanCodeProcessor extends AbstractProcessor {
...
+      fileCreationErrors += error
+    })
+
+    if (fileCreationErrors != '') {
+      this.logger.error(`Could not create NOTICE file to determine scancode version ${fileCreationErrors}`)
+    }
+
+    this._versionPromise = execFile(`${this.options.installDir}/scancode`, ['--json-pp', '-', 'NOTICE'])
```
nellshamrell, line 124: This method of getting the version was suggested by @pombredanne in comments on #459

test/unit/providers/process/scancodeTests.js
```diff
@@ -98,6 +102,12 @@ describe('ScanCode process', () => {
   })
 
   afterEach(function () {
+    // Check if NOTICE file (generated when checking the ScanCode version) exists
+    // If it does, delete it
+    if (fs.existsSync('NOTICE')) {
```
nellshamrell, line 107: remove the NOTICE file after it is generated when the test is run (I'm open to ideas on how to stub or mock this instead)

test/unit/providers/process/scancodeTests.js
```diff
@@ -52,7 +54,7 @@ describe('ScanCode process', () => {
   it('should handle gems', async () => {
     const { request, processor } = setup('2.9.8/gem.json')
     await processor.handle(request)
-    expect(request.document._metadata.toolVersion).to.equal('1.2.0')
+    expect(request.document._metadata.toolVersion).to.equal('30.1.0')
```
disulliv, line 57: Was this test case changed somewhere?

DevDockerfile
```diff
@@ -8,7 +8,7 @@ ARG BUILD_NUMBER=0
 ENV CRAWLER_BUILD_NUMBER=$BUILD_NUMBER
 
 # Ruby and Python Dependencies
-RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests curl bzip2 build-essential libssl-dev libreadline-dev zlib1g-dev cmake python3 python3-dev python3-pip && \
+RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests curl bzip2 build-essential libssl-dev libreadline-dev zlib1g-dev cmake python3 python3-dev python3-pip xz-utils libxml2-dev libxslt1-dev libpopt0 && \
```
disulliv, line 11: Are all of these new dependencies needed?

test/unit/providers/process/scancodeTests.js
disulliv: I'm not as familiar with best practices for functionality like this, but there might be a better way? Perhaps waiting for confirmation from someone else before merging would be best here.

DevDockerfile
nellshamrell: They are, according to @pombredanne in #459

test/unit/providers/process/scancodeTests.js
nellshamrell: I think there is - I'm going to see if I can stub out the file creation somehow, which will feel much less icky.

test/unit/providers/process/scancodeTests.js
nellshamrell: It was when I changed the result of the stubbed method (in line #90) - I suppose I could have left it the same, but it felt better make the stubbed value be the same as the current version of ScanCode.

test/unit/providers/process/scancodeTests.js
nellshamrell: It might make more sense to (in providers/process/scancode.js) create the NOTICE file with the fs module, rather than using child_process for it. This also may be easier to stub and test. Trying that out :)

nellshamrell: @disulliv done!

pombredanne: Actually, sorry if I might not have understood your use case correctly... I thought you wanted to get the version from a scan (which comes in the headers). To get it in general your previous approach to parse the --version output was fine.
And I created an upstream issue to make it even easier to parse: https://github.com/nexB/scancode-toolkit/issues/2856

nellshamrell: Ah, cool! I will change it back :)

test/unit/providers/process/scancodeTests.js
nellshamrell: Resolving - I reverted back to the way we used to determine the ScanCode version (and back to the original tests)

test/unit/providers/process/scancodeTests.js
nellshamrell: Done!

providers/process/scancode.js
nellshamrell: This was reverted back to how we determined the version in #459

providers/process/scancode.js
nellshamrell: This was reverted back to how we determined the version in https://github.com/clearlydefined/crawler/pull/459

disulliv (approved):

Issue #464, opened 2022-02-17
Reduce the number of fetches when harvesting one package


qtomlinson:
Package fetch from repository is triggered multiple times during harvesting one package.  The fetch is currently dispatched once per processor (e.g. licensee or scancode).  A package is processed by 3 processors at the moment, resulting in 3 round trips to the repository.  Presumably, only one fetch from the repository is necessary.

qtomlinson:
Sample log from processing a maven package: 3 maven fetches were dispatched.  
```
[V] Finished getRequest (6ms) package@cd:/maven/gradleplugin/io.github.lognet/grpc-spring-boot-starter-gradle-plugin/4.5.7:mM {"crawlerId":"979f84c6-3adc-4cd8-97e6-cb5692a96377","buildNumber":"0"}
[V] Started fetching package@cd:/maven/gradleplugin/io.github.lognet/grpc-spring-boot-starter-gradle-plugin/4.5.7:mM {"loopName":"0","cid":"14","root":"self","crawlerId":"979f84c6-3adc-4cd8-97e6-cb5692a96377","buildNumber":"0"}
[V] Finished fetching (2ms) package@cd:/maven/gradleplugin/io.github.lognet/grpc-spring-boot-starter-gradle-plugin/4.5.7:mM {"loopName":"0","cid":"14","root":"self","crawlerId":"979f84c6-3adc-4cd8-97e6-cb5692a96377","buildNumber":"0"}
[V] Started processing package@cd:/maven/gradleplugin/io.github.lognet/grpc-spring-boot-starter-gradle-plugin/4.5.7:mM {"loopName":"0","cid":"14","root":"self","crawlerId":"979f84c6-3adc-4cd8-97e6-cb5692a96377","buildNumber":"0"}
[V] Finished processing (3ms) package@cd:/maven/gradleplugin/io.github.lognet/grpc-spring-boot-starter-gradle-plugin/4.5.7:mM {"loopName":"0","cid":"14","root":"self","crawlerId":"979f84c6-3adc-4cd8-97e6-cb5692a96377","buildNumber":"0"}
[I] Traversed package@cd:/maven/gradleplugin/io.github.lognet/grpc-spring-boot-starter-gradle-plugin/4.5.7  {"loopName":"0","cid":"14","root":"self","outcome":"Traversed","time":12,"crawlerId":"979f84c6-3adc-4cd8-97e6-cb5692a96377","buildNumber":"0"}
[...]

PR #467, opened 2022-03-04
Reduce the number of fetches harvesting one package

qtomlinson:
Implemented FetchResult and a cache of FetchResults in the Dispatcher to prevent multiple subsequent fetches for the same coordinates.  The fetch result is currently cached for one day and can be reused for various code analyses: licensee, scancode, fosology, and, in the future, "reuse".
Also implemented a cache for inProgressFetches (promises) to avoid multiple concurrent fetches for the same coordinates.

Task: https://github.com/clearlydefined/crawler/issues/464


MichaelTsengLZ:
Hi @qtomlinson, I will review this PR next Monday since I have some small tasks that need to fix at this point.

MichaelTsengLZ:
> Hi @qtomlinson, I will review this PR next Monday since I have some small tasks that need to fix at this point.

Looks like I need some more time to review it.

MichaelTsengLZ:
Hi @qtomlinson, I take a quick look at this PR and everything looks good to me. 
There is just one question. This cache is a local cache and won't share across multiple crawlers. Therefore, the caching may be less hit. Have you considered this?

MichaelTsengLZ:
@qtomlinson  looks like you are right. I just confirmed the code in the screenshot.

image

MichaelTsengLZ (approved):

qtomlinson:
@MichaelTsengLZ See also AbstractProcessor.linkAndQueueTool and request.queueRequests.

MichaelTsengLZ:
> @MichaelTsengLZ See also AbstractProcessor.linkAndQueueTool and request.queueRequests.

I double-checked these two `codeAbstractProcessor.linkAndQueueTool` and `request.queueRequests` and here in the [dockerfile](https://github.com/clearlydefined/crawler/blob/f49495d09eaa29ddb63d4c472f057a02fa05032c/Dockerfile#L69), I confirmed that we use an Azure storage queue instead of a memory queue. So, I think all the requests are shared across all the crawlers. Those crawlers pop messages from the azure storage queue and push messages to the azure storage queue.

MichaelTsengLZ:
> Hi @qtomlinson, I take a quick look at this PR and everything looks good to me. There is just one question. This cache is a local cache and won't share across multiple crawlers. Therefore, the caching may be less hit. Have you considered this?

I double-checked these two function `codeAbstractProcessor.linkAndQueueTool` and `request.queueRequests`. And I double checked [dockerfile](https://github.com/clearlydefined/crawler/blob/f49495d09eaa29ddb63d4c472f057a02fa05032c/Dockerfile#L69). 

I think we use an Azure storage queue instead of a memory queue on production/develop crawler. So, I think all the tasks/requests are shared across all the crawlers. Those crawlers pop messages from the azure storage queue and push messages to the azure storage queue. So the same coordinate may be spread to multiple crawlers to process.

qtomlinson:
Good Point!  Dev/Prod crawlers are configured differently than local dev environment.  In that case local cache is not as helpful.

qtomlinson:
@jeffmcaffer @MichaelTsengLZ ready for review

disulliv:
I'll take a closer look at the latest updates this week

qtomlinson:
To see the crawler shutdown sequence triggered, need to set init to true for crawler in docker-compose.yml.  See https://github.com/clearlydefined/docker_dev_env_experiment/pull/7  

For production and development builds: 
1. ENTRYPOINT need to be changed to ["node", "index.js"] in Dockerfile, and DevDockerfile
2. use --init in the docker run command
See https://maximorlov.com/process-signals-inside-docker-containers/

disulliv (approved):
Changes look good to me, super thorough testing too! 

> ENTRYPOINT need to be changed to ["node", "index.js"] in Dockerfile, and DevDockerfile
> use --init in the docker run command
Could you include these changes as well? Thanks!

qtomlinson:
@disulliv updated Dockerfile, DevDockerfile and dev scripts.  See https://github.com/clearlydefined/crawler/pull/467/commits/e847fc3affe17f102a01687e0a8c17bee7238407.  Are there scripts for dev and production deployment?  Need to consider the following two points for those:
1. Use docker stop (instead of docker kill) so that the graceful shutdown of service can be completed.
2. Use docker run --init (as in the dev scripts)

disulliv:
I don't think there's any deployment scripts. From what my investigation of the infra, the containers are pushed to a Docker and Azure Container Registry, then the crawler vms and/or app services pick up the latest container to be used.

It seems like for App Services, each instance needs to change the default startup command to call `docker run --init` instead oof the default `docker run`. This might require more community engagement since Clearly Defined / Microsoft doesn't own all the crawler instances. I think Google has a set and maybe others we don't know about.

What would happen if `--init` isn't used?

qtomlinson:
@disulliv During the graceful shutdown of the crawler service, requests on the crawler's local queueset are published to the shared global queueset so that other crawler instances can continue processing them.  When --init is not used, the cralwer service will not receive the shutdown signal and the service graceful shutdown will not be triggered.  The consequence is that the last package harvests will be partial (the related requests are on the local in memory queueset, which are not published to global queue set and therefore lost after shutdown).  The max number of affected package harvest is determined by the cdConfig.crawler.counter (currently set at 2).

Issue #450, opened 2021-12-09
Possible problem with harvesting release dates for pods and maven google packages


nellshamrell:
Looks like we're not getting the release date when we harvest pods or maven google packages

* https://dev.clearlydefined.io/definitions/pod/cocoapods/-/SwiftCommonTools/0.1.2
* https://clearlydefined.io/definitions/maven/mavengoogle/android.arch.lifecycle/viewmodel/1.1.0

qtomlinson:
For https://clearlydefined.io/definitions/maven/mavengoogle/android.arch.lifecycle/viewmodel/1.1.0, the decompressed archive does not contain either META-INF/MANIFEST.MF or META-INF/maven/android.arch.lifecycle/viewmodel/pom.properties.  These are the two files currently used to infer release date.

PR #469, opened 2022-03-08
Fixes for maven and pod release dates

qtomlinson:
Two different bugs here:
1. The maven google artifact of interest does not have specific files that are used to infer dates (Details see https://github.com/clearlydefined/crawler/issues/450#issuecomment-1061287307).  Added fallback to check for stat of any file in the decompressed folder.
2. Release date needs to be included in the document for pods (PodExtract).  This is later used by service layer to extract date information.

Task: https://github.com/clearlydefined/crawler/issues/450


qtomlinson:
@jeffwilcox Ready for review.  Crawler side of the fix.

jeffwilcox (approved):

PR #474, opened 2022-05-27
[WIP]: Build failed when creating docker image.

MichaelTsengLZ:
Background:

[This](https://dev.azure.com/clearlydefined/ClearlyDefined/_build/results?buildId=8625&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=457762fb-7913-5721-db09-27cc274cbeba&l=4230) build failed because of the following error.

```
ERROR:  Error installing licensee:
	The last version of faraday (< 3, >= 1) to support your Ruby & RubyGems was 1.10.0. Try installing it with `gem install faraday -v 1.10.0` and then running the current command again
	faraday requires Ruby version >= 2.6. The current ruby version is 2.5.0.
```

The [faraday@2.3.0](https://rubygems.org/gems/licensee/versions/9.12.0/dependencies) package is a run time dependency of licensee@9.12.0. And it requires RubyGem >= 2.6, [here](https://rubygems.org/gems/faraday/versions/2.3.0).

Changes:
1. Update RubyGem to 3.1.2 since ruby-build mark it as the latest stable release.
2. Remove `gem install nokogiri:1.12.5` because licensee@9.12.0 will install nokogiri:1.13.6 which requires RubyGem >= 2.6.0
3. gem options `--no-rdoc` and `--no-ri` are replaced with `--no-document`

### DIFF
--- a/DevDockerfile
+++ b/DevDockerfile
@@ -10,10 +10,10 @@ ENV CRAWLER_BUILD_NUMBER=$BUILD_NUMBER
 # Ruby and Python Dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends --no-install-suggests curl bzip2 build-essential libssl-dev libreadline-dev zlib1g-dev cmake python3 python3-dev python3-pip xz-utils libxml2-dev libxslt1-dev libpopt0 && \
   rm -rf /var/lib/apt/lists/* && \
-  curl -L https://github.com/rbenv/ruby-build/archive/v20180822.tar.gz | tar -zxvf - -C /tmp/ && \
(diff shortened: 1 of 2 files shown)


MichaelTsengLZ:
After upgrading RubyGem to 3.1.2, licensee:9.12.0 starts to have hard time to find license in LICENSE file, like component npm/npmjs/-/caniuse-lite/1.0.30001344. So we pin to the previous version of nokogiri and faraday instead of upgrade RubyGem.

DragosDanielBoia (approved):

PR #476, opened 2022-06-21
Parse declared license from pkg.go.dev html

MichaelTsengLZ:
1. The docker file uses node:16 now. So update the package-lock.json to v2
2. Parse declared license from pkg.go.dev HTML. Because each crawler instance has its own virtual IP and each of them will have 2 concurrent calls to pkg.go.dev. It's OK NOT to implement a mechanism to control all the crawler calls to pkg.go.dev for now.


qtomlinson:
hm... parsing license based on class names or id names in a html page can be fragile.  Unlike a REST api, these class (or id) names are not documented and therefore not contractual, and can be changed at anytime.  Whenever the go repository ui page is upgraded, this can be broken.  
When implementing gradle plugin support in the crawler, I had a similar discussion with Nell regarding parsing html page for the official release date for gradle plugin (no repository api available).  Both of us decided against it.

MichaelTsengLZ:
There are two reasons why I think it’s OK. 
 
1. If the license parse failed, it won’t affect any following process(I have a PR later to fix this). We try our best to get the declared license, if we could not, then skip it.
2. For pkg.go.dev license HTML is open source. https://github.com/golang/pkgsite/blob/master/static/frontend/unit/licenses/licenses.tmpl. It will change but we will know how it’s been changed.
3. An alert will be set up if the license HTML has been changed.
4. There is an ongoing issue asking for a REST API for pkg.go.dev. https://github.com/golang/go/issues/36785. We could replace the HTML parse later with the API call.

What do you think?

qtomlinson:
Sounds good if replacement with the future API call (when that becomes available) is in the plan.

MichaelTsengLZ:
Cool. Thanks.

PR #477, opened 2022-06-23
Doesn't throw error when html changed

MichaelTsengLZ:


haydenk82 (approved):

qtomlinson (approved):

providers/fetch/goFetch.js
```diff
@@ -56,7 +57,6 @@ class GoFetch extends AbstractFetch {
       if (err instanceof DeferError) {
         return request.markRequeue('Throttled', err.message)
```
qtomlinson, line 58: Request can only be retried (requeue) to a max of number of times (See crawler._requeue).  May not want to fail the fetch for retried max number of times and not able to retrieve license data here.

MichaelTsengLZ: Good catch. Here is the PR to fix this. https://github.com/clearlydefined/crawler/pull/478

PR #478, opened 2022-06-24
Do not requeue after get requeued 5 times for Go

MichaelTsengLZ:


providers/fetch/goFetch.js
```diff
@@ -54,7 +54,8 @@ class GoFetch extends AbstractFetch {
     try {
       registryData = await this._getRegistryData(spec)
     } catch (err) {
-      if (err instanceof DeferError) {
+      const maxRequeueCnt = 5
```
disulliv, line 57: Should this be defined as an environment variable instead?

disulliv (approved):

qtomlinson (approved):

providers/fetch/goFetch.js
qtomlinson: For a reduced scope, can add a descriptive constant set to value of 5 (instead of environment variable for now) in cdConfig.js, and pass the value in crawler (similar to crawler.count) and into fetch.go.   This will make the code a bit more maintainable.

qtomlinson (approved):

PR #480, opened 2022-07-28
Fixed error creating scopedQueueSets

qtomlinson:
When creating scopedQueueSets with non memory global queues, there is error.
Fixed by passing a copy of global queue options for memory based local queue sets.


qtomlinson:
@MichaelTsengLZ ready for review

MichaelTsengLZ (approved):

PR #493, opened 2022-10-20
Delay cleanup of cached fetch result and add StorageBackedInMemoryQueue

qtomlinson:
Added delay cleanup of the cached fetch result if it is in use.  
Also added StorageBackedInMemoryQueue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
   },
   "devDependencies": {
     "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
     "chai-spies": "^1.0.0",
     "eslint": "^7.5",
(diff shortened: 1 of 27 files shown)


qtomlinson:
@mpcen ready for review

mpcen (approved):

package.json
```diff
@@ -78,6 +78,7 @@
   },
   "devDependencies": {
     "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
```
mpcen, line 81: wondering why the lock file changes weren't committed

qtomlinson: good point! will fix that.

Issue #494, opened 2022-11-14
"stdout maxBuffer length exceeded" in reuse processor when processing large packages


qtomlinson:
The following error messages were observed in the container log of dev deployment:
"2022-11-07T16:39:42.265152221Z [I] Error reuse@cd:/git/github/dotnet/runtime/3a25a7f1cc446b60678ed25c9d829420d6321eba stdout maxBuffer length exceeded {"loopName":"0","cid":"sf","root":"component@cd:/nuget/nuget/-/microsoft.extensions.configuration/6.0.1","gitSize":716648,"outcome":"Error","time":18004111,"crawlerId":"58d5c85f-7aa0-4ce2-aca4-7af1b34c9a59","crawlerHost":"dev","buildNumber":"20221014.1"}"

qtomlinson:
Additional test cases:
git/github/swc-project/swc/da1fd033b846696dec555d2e7cfb2b62b0f4d7eb
git/github/babel/babel/7b7ab94066448cd495b33b51512f109589d63fd1
git/github/babel/babel/9a4269e7b6bc8cf2b623de2547b75cc541b757ca

PR #495, opened 2022-11-14
Fix "stdout maxBuffer length exceeded" error in reuse processor

qtomlinson:
Current reuse processing utilizes stdout with a predetermined buffer size. The buffer size is not enough for some large packages. Use the file output to overcome the buffer size limit.

Command 'docker run --rm --volume $(pwd):/data fsfe/reuse spdx --help' shows --output or -o can be used to specify an output file

Additional documentation:
-reuse documentation
(https://reuse.readthedocs.io/en/stable/readme.html#run-in-docker) 
-reuse codebase
https://git.fsfe.org/reuse/tool/src/commit/2b0d470919f98625f31b64819c889b0c54fc0180/src/reuse/spdx.py#L27

Task: https://github.com/clearlydefined/crawler/issues/494

### DIFF
--- a/providers/process/fsfeReuse.js
+++ b/providers/process/fsfeReuse.js
@@ -5,7 +5,7 @@ const AbstractProcessor = require('./abstractProcessor')
 const { promisify } = require('util')
 const execFile = promisify(require('child_process').execFile)
 const { merge } = require('lodash')
-const { readdirSync } = require('fs')
+const { readdirSync, promises: { readFile } } = require('fs')
 
 class FsfeReuseProcessor extends AbstractProcessor {
   constructor(options) {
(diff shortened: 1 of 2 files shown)


qtomlinson:
@mpcen @SebastianWolf-SAP ready for review :)

mpcen (approved):

SebastianWolf-SAP (approved):
I remember the issues with the buffer, but was unsure if I should use a file (maybe there would be other drawbacks), so I used the buffer here... As @mpcen approves the change, of course you get my thumbs up as well.

qtomlinson:
@SebastianWolf-SAP Thanks for feedback! Using file output adds the need for file cleanup.  The output file is cleaned up via the built-in temp file cleanup mechanism upon request processing complete.

PR #515, opened 2023-03-03
lowercasing package names for nuget api fetching

mpcen:
Calls to `https://api.nuget.org/v3-flatcontainer/${spec.name}/${spec.revision}/${spec.name}.nuspec` were failing when components contained uppercase chars thus marking them as "missing" when attempting to harvest.

This update lowercases all nuget component names before calling the nuget api

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,7 @@
   "editor.folding": false,
   "editor.tabSize": 2,
   "editor.detectIndentation": false,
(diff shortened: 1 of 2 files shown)


qtomlinson (approved):
The code change looks fine.  
Question: nugetFetch._getRegistryData also contains conversion to lower case for registry search.  If nuget package names are case insensitive, do we want to consider using entityCoordinates.toLowerCaseMap to normalize the input nuget package names?

PR #525, opened 2023-09-28
Exclude .git directory content when calculating package file count

qtomlinson:


qtomlinson:
@mpcen ready for review

qtomlinson:
@jeffwilcox @mpcen ready for review.

qtomlinson:
@elrayle  ready for review

qtomlinson:
The existing code only exclude .git directory itself, but not its containing files. This fix intends to exclude .git directory and its contents.

providers/process/abstractProcessor.js
```diff
@@ -150,15 +150,21 @@ class AbstractProcessor extends BaseHandler {
...
-      return !intersection(segments, exclusions).length
-    })
+    const filteredList = fullList.filter(file => this.isValidExcludingGit(file))
     return trimAllParents(filteredList, location).filter(x => x)
   }
 
+  _isValid(file, exclusions) {
+    if (!file) return false
+    const segments = file.split(/[\\/]/g)
+    return !intersection(segments, exclusions).length
```
elrayle, line 160: `intersection` can have false positives...

```
const filename = '.git/build.yaml'
const segments = filename.split(/[\\/]/g)
const intersectionCount = intersection(, [.git]).length // returns 1

const filename = 'build/.git'
const segments = filename.split(/[\\/]/g)
const intersectionCount = intersection(, [.git]).length // also returns 1
```

This may be unlikely for .git, but if this is reused for other paths, the behavior may not be what is expected.

The first question is should _isValid reject anything where the exclusion list appears anywhere in the path, which is what it does now?  Or should it reject anything that starts with an exclusion, which is the desired behavior to exclude the .git directory?

Either way, a comment on the function describing what to expect would be helpful to clear up any confusion.

elrayle (commented):
I like where you are going with this, especially pulling out the validation into a separate function.  I have a question about the expectation of the validation and whether or not the use of intersection could have false negatives.

providers/process/abstractProcessor.js
qtomlinson: Thanks for your comment. The .git directory is created by git operations, and its content is not part of the package content. So, ignoring .git and its contents if .git is present anywhere in the path makes sense. For example, it is possible to create a temp directory (e.g., /tmp/tempX) and then git clone into a subdirectory of tempX (e.g., /tmp/tempX/package), resulting in /tmp/tempX/package/.git. The original condition is to check the basename of a file (the last portion) and ignore it if it is '.git', resulting in ignoring the .git directory itself. For the contents of .git, .git would be in the middle of the path, and those files need to be excluded as well.

It is true that ignoring a directory and its contents if a specific directory is present anywhere in the path may not have the desired effect when considering other exclusions. I will refactor accordingly.

qtomlinson:
@lumaxis @elrayle ready for review

lib/utils.js
```diff
@@ -31,6 +32,13 @@ function trimAllParents(paths, parents) {
   return paths.map(path => trimParents(path, parents))
 }
 
+function isValidExcludingGit(file) {
+  const exclusions = ['.git']
+  if (!file) return false
+  const segments = file.split(/[\\/]/g)
+  return !intersection(segments, exclusions).length
```
lumaxis, line 39: To me, this would make the intention of this logic a little bit clearer than using the exclamation mark operator 🤔 
```suggestion
  return intersection(segments, exclusions).length === 0
```

lib/utils.js
```diff
@@ -31,6 +32,13 @@ function trimAllParents(paths, parents) {
   return paths.map(path => trimParents(path, parents))
 }
 
+function isValidExcludingGit(file) {
```
lumaxis, line 35: Maybe this would be a little easier to understand what the function does?

```suggestion
function isFileValidAndNotGitRelated(file) {
```

elrayle (commented):
Good to see this bug fixed.  Really nice to see a function created that can be called from multiple places resulting in the same behavior.

lib/utils.js
lumaxis: `isFileValidAndExcludesGit` or `isFileValidAndUnrelatedToGit` maybe?

lumaxis: As we discussed today with @elrayle, naming this `isGitFile` or similar and using the result appropriately at the call site is probably the cleanest solution 👌🏼

providers/process/abstractProcessor.js
```diff
@@ -150,7 +150,7 @@ class AbstractProcessor extends BaseHandler {
    */
   async filterFiles(location) {
     const fullList = await this.getFiles(location)
-    const filteredList = fullList.filter(file => isValidExcludingGit(file))
+    const filteredList = fullList.filter(file => file && !isGitFile(file))
```
lumaxis, line 153: We don't necessarily need to check if file is `undefined` here since `isGitFile` already does: `if (!file) return false`

lumaxis (approved):

providers/process/abstractProcessor.js
qtomlinson: Hm... I am afraid that the check for undefined file is still necessary.  In case of falsy file, isGitFile(file) returns false, and !isGitFile returns true.  Using !isGitFile alone would not exclude falsy file values.

PR #526, opened 2023-10-11
Update readme

qtomlinson:


README.md
```diff
@@ -80,6 +80,18 @@ Process the source, if any:
 
 The crawler's output is stored for use by the rest of the ClearlyDefined infrastructure -- it is not intended to be used directly by humans. Note that each tool's output is stored separately and the results of processing the component and the component source are also separated.
 
+### More on `type`...
```
elrayle, line 83: It would be helpful to have a pointer to this section from the place where `type` is being described on line 55.  I can't make a suggestion outside the area that is being changed, so I'll add a suggested rewrite here.

Original
> The request `type` describes the crawling activity being requested. For example, "do `package` crawling". It is typically the same as the `type` in the url (see below). There are some more advanced scenarios where the two values are different but for starters, treat them as the same. The general form of a request URL is (note: it is a URL because of the underlying crawling infrastructure, the `cd` scheme is not particularly relevant)

Proposed
> The request `type` describes the crawling activity being requested. For example, "do `package` crawling" (see [More on `type`](https://github.com/clearlydefined/crawler/blob/master/README.md#more-on-type) for a description of valid type values). It is typically the same as the `type` in the url (see segments description below). There are some more advanced scenarios where the two values are different but for starters, treat them as the same. The general form of a request URL is (note: it is a URL because of the underlying crawling infrastructure, the `cd` scheme is not particularly relevant)

elrayle, line 83: Since none of the other sections use `...` in headers and it makes it easier to create the internal reference for linking to this section, I'd recommend the following change.

```suggestion
### More on `type`
```

README.md
```diff
@@ -80,6 +80,18 @@ Process the source, if any:
...
+The `type` in the request object typically corresponds to a internal processor in CD.
```
elrayle, line 84:
```suggestion
The `type` in the request object typically corresponds to an internal processor in CD.
```

elrayle (approved):
This is excellent and will be very helpful for those new to ClearlyDefined.  Thanks for adding to the docs!

README.md
qtomlinson: Good suggestion!

qtomlinson:
Thanks for the feedback!
@jeffwilcox @elrayle ready for review

qtomlinson:
@jeffwilcox @mpcen Ready for review. Any feedback will be very much appreciated.  How do we go about getting this merged? 🙂

PR #537, opened 2024-01-18
Update scancode-toolkit to latest version

lumaxis:
This is the initial version of the changes needed to update [scancode-toolkit](https://github.com/nexB/scancode-toolkit) to the latest version. Further changes will be necessary in [`clearlydefined/service`](https://github.com/clearlydefined/service) to be able to handle the output from recent versions of ScanCode where the data format has changed fairly dramatically in parts.


config/cdConfig.js
```diff
@@ -72,20 +72,21 @@ module.exports = {
...
         '--generated',
+        '--license',
+        '--license-clarity-score',
+        '--license-references',
+        '--license-text',
+        '--license-text-diagnostics',
         '--summary',
-        '--summary-key-files'
+        '--tallies',
+        '--tallies-key-files'
```
lumaxis, line 89: This is my understanding of what we need to get all the necessary data from ScanCode.
As I understand, `--tallies-key-files` is the successor in spirit to `--summary-key-files` and requires `--tallies`. The rest is just reordered and `--is-license-text` removed, since it is no longer supported.

Dockerfile
```diff
@@ -85,7 +53,7 @@ RUN git config --global --add safe.directory '*'
 
 COPY package*.json /tmp/
 COPY patches /tmp/patches
-RUN cd /tmp && npm install --production
+RUN cd /tmp && npm install --omit=dev
```
lumaxis, line 56: `--production` has been deprecated in `npm`, the new way of doing this is by using `--omit=`.

.vscode/launch.json
```diff
@@ -45,21 +57,20 @@
...
       "env": {
         "NODE_ENV": "localhost"
       },
       "console": "internalConsole"
-    },
-    {
-      "type": "node",
-      "request": "launch",
-      "name": "Launch in WSL",
-      "useWSL": true,
```
lumaxis, line 61: This is not supported in VSCode anymore and has therefore been removed

DevDockerfile
```diff
@@ -57,4 +57,4 @@ ENV NODE_ENV "localhost"
 
 ENV PORT 5000
 EXPOSE 5000
-ENTRYPOINT ["node", "index.js"]
+CMD ["node", "index.js"]
```
qtomlinson, line 60: --entrypoint is overriden in Readme.doc and debug.bat.  may need update.

Dockerfile
```diff
@@ -1,14 +1,8 @@
 # Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 # SPDX-License-Identifier: MIT
 
-#FROM fossology/fossology:3.4.0 as fossology
-#COPY fossology_init.sh fossology_init.sh
-#RUN ./fossology_init.sh
-
 FROM node:18-bullseye
 ENV APPDIR=/opt/service
```
qtomlinson, line 5: Question: Is it possible to use node:18 for crawler after this ScanCode update?  If so, this can be done as a separate PR in the future.

qtomlinson (commented):
Thanks for the detailed explanation!

qtomlinson (approved):

Dockerfile
elrayle: I'd be curious if changing to node:18 requires additional changes for this to work.  If we can change here, then we'd want to change in dev too.  I agree that this is not required to be part of this PR.

Dockerfile
```diff
@@ -1,14 +1,8 @@
 # Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 # SPDX-License-Identifier: MIT
 
-#FROM fossology/fossology:3.4.0 as fossology
-#COPY fossology_init.sh fossology_init.sh
-#RUN ./fossology_init.sh
```
elrayle, line 6: Thanks for helping to keep files clean by removing commented out code!

elrayle (approved):
This all looks good to me.  I see a couple of outstanding questions from @qtomlinson.  Once those are resolved, let's get this merged.

Thanks for all the great work that went into getting scancode upgraded and adding LicenseRefs support.

Dockerfile
lumaxis: It unfortunately doesn't work out of the box but might do with the changes from https://github.com/clearlydefined/crawler/pull/549 ?

qtomlinson: Thanks for checking!  Upgrade to node:18 is a separate task that we can look into after this PR gets merged.

DevDockerfile
lumaxis: Decided to undo this change after all and just stick with what's there for now.

PR #538, opened 2024-01-18
Improve formatting and linting setup

lumaxis:
I stumbled over the confusingly leftover `.prettierrc` and took a quick look at the ESLint setup. It's now updated to the latest version and the config updated to allow ECMA version syntax and features that should closely match Node 18.


qtomlinson:
With eslint deprecate formatting rules, it maybe worthwhile to keep a formatter in place for consistent formatting in the code base.  Any thoughts?

lumaxis:
@qtomlinson @elrayle I've taken on the suggestion to keep Prettier and improved the project setup so it is properly integrated now and e.g. runs automatically in CI and has good editor integration with VS Code.

The PR now also includes a commit that fixes all previous issues that Prettier fixes, so it might be easier to review this commit by commit.

Additionally, I'd suggest to only merge this after #532, so that that doesn't have to resolve a bunch of formatting issues.

Let me know what you think! If we like this setup, I'd be happy to port it to the service and website repos eventually, too.

package.json
```diff
@@ -5,10 +5,16 @@
   "main": "./index.js",
   "scripts": {
     "start": "node ./index.js",
-    "test": "nyc mocha \"test/unit/**/*.js\" --timeout 20000 && eslint .",
+    "test": "npm run mocha && npm run lint",
+    "mocha": "nyc mocha \"test/unit/**/*.js\"",
```
qtomlinson, line 9: shall we keep the time out when running mocha?

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "printWidth": 120,
+  "singleQuote": true,
+  "semi": false
```
qtomlinson, line 4: Shall we add the following config to be consistent with the current style in the code base and minimize the formatting fixes?
"trailingComma" : "none",
"arrowParens": "avoid"

https://prettier.io/docs/en/options.html#trailing-commas
https://prettier.io/docs/en/options.html#arrow-function-parentheses

package.json
qtomlinson: This will help to improve consistency in the code base.  Question: Is it better to separate the fixing formatting in a separate PR?

qtomlinson:
This will definitely help to ensure consistency in the code base.  Thanks for putting this together!

package.json
lumaxis: I'd like to leave it out if possible as it can potentially hide issues. 20 seconds seems very long and I have not been able to trigger any timeouts when testing this change locally. Plus, the CI builds seem to work fine, too. I'd suggest to try it without it and re-add a as small as possible timeout if it becomes necessary through e.g. flaky tests.

.prettierrc.json
lumaxis: Sure! I don't mind either way, generally, it seems to be encouraged to run Prettier with as few exceptions as possible.

@qtomlinson Do you know if there ever has been any other codified style guide for ClearlyDefined projects' code?

package.json
qtomlinson: sounds good

.prettierrc.json
qtomlinson: There is guide on formatting and coding at https://docs.clearlydefined.io/contributing-code.  Point 5 in `Formatting` aligns with "arrowParens": "avoid"

lumaxis: Great, that's good to know!

I added both configurations you suggested. The number of changed files shrunk a bit but is still fairly large since Prettier hasn't been run in such a long time. I think this is good to go from my perspective!

qtomlinson (approved):

lumaxis:
As mentioned above, I'll hold off merging this until https://github.com/clearlydefined/crawler/pull/532 is in to avoid any conflict issues 🙏🏼

PR #551, opened 2024-01-31
Add GitHub Actions workflow to run tests

lumaxis:
As discussed in today's dev sync, this paves our path for consolidating build & tests on GitHub Actions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
           cache: 'npm'
 
       - name: Install dependencies
-        run: npm install
+        run: npm ci
 
       - name: Run tests
         run: npm test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,29 @@
+name: Run tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.1
+
+      - uses: actions/setup-node@v4.0.1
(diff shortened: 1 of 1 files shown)


elrayle:
How does this compare to the [service repo action](https://github.com/clearlydefined/service/blob/master/.github/workflows/build.yml) that was put in several months ago to build and run tests?  I'd like to standardize these as much as possible across the major repos.

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
+name: Run tests
```
elrayle, line 1: I'd suggest the following name.  If the PR fails this action, it can be because the build failed or because tests failed.

```suggestion
name: Build and Run Tests
```

elrayle (approved):
This looks good.  It will be nice to have a `Build and Run Tests` action in all the major repos.

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v4.1.1
+
+      - uses: actions/setup-node@v4.0.1
+        with:
+          node-version: 18
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm install
```
qtomlinson, line 26: `npm ci`  instead?  it is meant to be used for automated environments

.github/workflows/test.yml
lumaxis: Sure, I don't mind but I left this off initially because we don't really run any build right now 🤔

qtomlinson (approved):

.github/workflows/test.yml
lumaxis: Spoke about this offline, since this doesn't actually run a build today, going to keep as is.

PR #558, opened 2024-02-15
Remove rimraf

lumaxis:
Updating `rimraf` to the latest version requires a small code change due to the change in how the library handles its exports (there is no default export anymore), plus it now supports Promises natively by default, hence we don't need to manually resolve and reject anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "promise-retry": "^1.1.1",
         "request-promise-native": "^1.0.5",
         "requestretry": "^4.0.0",
-        "rimraf": "^3.0.0",
+        "rimraf": "^5.0.5",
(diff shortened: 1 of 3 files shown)


qtomlinson:
Since node 14, there is [fsPromises.rm](https://nodejs.org/api/fs.html#fspromisesrmpath-options).  "To get a behavior similar to the rm -rf Unix command, use [fsPromises.rm()](https://nodejs.org/api/fs.html#fspromisesrmpath-options) with options { recursive: true, force: true }."  Might be better in terms of future maintenance?

lumaxis:
@qtomlinson Nice! Much easier 🙂

qtomlinson (approved):

PR #560, opened 2024-02-22
Update README.md

yashkohli88:
The README.md file has been updated to fix the setup issues in development environment on MAC systems.

Platform flag has been added in docker build and docker run commands while building/pulling the crawler image. It has been set to the value "**linux/amd64**" in order to maintain the compatibility.


lumaxis (approved):
Alternatively, people could use `export DOCKER_DEFAULT_PLATFORM=linux/amd64` 👍🏼

PR #566, opened 2024-04-10
Update PodExtract tool version

qtomlinson:
1. Add unit tests for verification purposes.
2. Refactor toolVersion in PodExtract. Reasons in [commit message](https://github.com/clearlydefined/crawler/commit/794fef2d4bef3ae8f2e139d29c528d251b9dbfda)
3. The recent fix to exclude content in the .git directory (https://github.com/clearlydefined/crawler/pull/525) from pod packages will cause the file count to be different from the previous version. Update the toolVersion for PodExtract to 2.0.0 to reflect this.


lumaxis (approved):
Thanks for the detailed commit messages to help understand context!

PR #570, opened 2024-04-17
Rename variable for consistency

lumaxis:
This variable is called `providerMap` in all other `*Fetch.js` implementations. Renaming this here makes the name consistent across files.


qtomlinson (approved):
Thank you for the quick fix!

PR #574, opened 2024-05-01
add sha and version to ‘/‘ endpoint

elrayle:
### Description

Add configurations for `BUILD_SHA` and `APP_VERSION` to use for debugging.

### Related Work

* https://github.com/clearlydefined/operations/pull/94
* https://github.com/clearlydefined/service/pull/1199

### Unresolved

The deploy isn't happening through the reusable workflows that would set these configs.  Additionally, it is not clear how deploys are happening in the current Azure DevOps deploy.  Some investigation is needed before merging.

### DIFF
--- a/.github/workflows/build-and-deploy-dev.yml
+++ b/.github/workflows/build-and-deploy-dev.yml
@@ -23,7 +23,7 @@ jobs:
   build-and-deploy:
     name: Build and Deploy
     needs: upload-package-lock-json
-    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@0a6c2607470e2e0e49bf2d6636223bcd62c1a7f8
+    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v3.0.0
     secrets: 
       AZURE_CREDENTIALS: ${{ secrets.AZURE_CREDENTIALS }}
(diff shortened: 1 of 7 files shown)


ghcrawler/routes/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = router
+
+let version
+let sha
+function setup(buildsha, appVersion) {
+  version = appVersion
+  sha = buildsha
+  return router
+}
```
lumaxis, line 19: This works but I'm not sure how well I like keeping the values as global state in this file. Couldn't we just access the config directly here instead of passing as variables?
Additionally, I'm wondering if it'd be worth to still escape these values, even though they should usually be under our control.

.github/workflows/build-and-deploy-dev.yml
```diff
@@ -23,7 +23,7 @@ jobs:
   build-and-deploy:
     name: Build and Deploy
     needs: upload-package-lock-json
-    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@v2.0.0
+    uses: clearlydefined/operations/.github/workflows/app-build-and-deploy.yml@0a6c2607470e2e0e49bf2d6636223bcd62c1a7f8
```
ljones140, line 26: Needs to be changed once operations PR merged

RomanIakovlev (approved):

Issue #568, opened 2024-04-13
Test Conda components harvest


qtomlinson:
This documents the QA work that is required for the [implementation of Conda component harvest](https://github.com/clearlydefined/crawler/issues/535):

- Trigger harvest and verify that component is harvest correctly
case 1. coordinates with version specified,
case 2. latest version of component when version is not specified
- related source component correctlly identified and harvested

- Verify definition correctly computed:
case 1, license detected in scancode package information.
case 2 if license not detected in scancode package level, figure out license based on license files in /info directory.

- Add integration test
add a test case (coordinates) in testConfig.js + fixture for Conda in operation repo

qtomlinson:
@lisahoong Please feel free to add on the cases that you can think of.  Appreciate your help!

qtomlinson:
Following our code review discussion, I would like to suggest testing the impact of implementing [caching indexes](https://github.com/clearlydefined/service/blob/82141499d431ffd665002a75c0b981d2a25463e3/routes/originConda.js#L16) (specifically repodata.json and channeldata.json) on the service's performance. This can be done by calling the /originConda endpoint to query Conda components.

qtomlinson:
[Integration Test Readme](https://github.com/clearlydefined/operations/tree/main/tools/integration) can be helpful.  Please also feel free to put up a PR to enhance it for others.  [Sample PR](https://github.com/clearlydefined/operations/pull/70) might be helpful for adding an integration test case.  All the test components are tested in definitionTest.js and attachmentTest.js.  When there are attachments in your test definition, more fixtures may need to be added and loaded for the attachmentTest.js.

lamarrr:
the tests have been added to https://github.com/clearlydefined/operations/pull/72/files and verified

qtomlinson:
> Following our code review discussion, I would like to suggest testing the impact of implementing [caching indexes](https://github.com/clearlydefined/service/blob/82141499d431ffd665002a75c0b981d2a25463e3/routes/originConda.js#L16) (specifically repodata.json and channeldata.json) on the service's performance. 

@lamarrr any update?

lamarrr:
I did send the memory impact of the crawler via email yesterday.
The service has a similar memory increase impact. perhaps you need it documented somewhere?
There are no existing tests to measure it as I had to look at the overall memory consumption of the crawler process in the docker container.

qtomlinson:
This is to document the APIs that can be used for test cases:
Post call to trigger harvest: 
[...]

PR #575, opened 2024-05-07
Fix ENOENT error during harvesting Conda components

qtomlinson:
This is to fix [ENONET exception](https://github.com/clearlydefined/crawler/issues/568#issuecomment-2098910877)  encountered during testing on dev deployment. It follows the same approach taken in [debianFetch.js](https://github.com/clearlydefined/crawler/blob/473a5603c8a149649c4de46f144dc9432ab452cf/providers/fetch/debianFetch.js#L40).

### DIFF
--- a/providers/fetch/condaFetch.js
+++ b/providers/fetch/condaFetch.js
@@ -11,7 +11,7 @@ const FetchResult = require('../../lib/fetchResult')
 class CondaFetch extends AbstractFetch {
   constructor(options) {
     super(options)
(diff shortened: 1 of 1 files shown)


lumaxis (approved):
Thanks! I wonder if we can do anything about either the env setup or the code to make this mechanism more resilient in the future so other contributors don't run into this discrepancy as easily 🤔

elrayle:
@lumaxis That refactor sounds like a great idea!

elrayle:
Thanks @qtomlinson for making this happen.

Issue #523, opened 2023-09-13
QUESTION: Rationale for classifiers taking precedence over license info


elrayle:
### Description

It appears that for pypi, classifiers take precedence over the license field when extracting the license information.  It is clear from the code how this is happening.  I'm wondering about the rationale for this approach.  Also if it is determined that one is correct and the other is not, is there a process for updating the license or the classifier as needed?

### Test

```
  it('parses the correct license information from classifiers in registry data', () => {
    const registryData = JSON.parse(fs.readFileSync('test/fixtures/pypi/registryData_lgpl2.json'))
    const declared = fetch._extractDeclaredLicense(registryData)
    expect(declared).to.be.equal('LGPL-2.0-only')
  })
```

### Fixture Data

[classifiers](https://github.com/clearlydefined/crawler/blob/a28303b6f729fbd2a61c38eb44c5f57402396bcc/test/fixtures/pypi/registryData_lgpl2.json#L6-L21)

```
    "classifiers": [
      ...
      "License :: OSI Approved :: GNU Lesser General Public License v2 (LGPLv2)",
      ...
    ],
```

[license](https://github.com/clearlydefined/crawler/blob/a28303b6f729fbd2a61c38eb44c5f57402396bcc/test/fixtures/pypi/registryData_lgpl2.json#L33)

```
    "license": "LGPL 2.1",
```

### Expected

With a specific license given in `info.license`, I would expect the license to be either `LGPL-2.0-or-later` or `LGPL-2.1-only`.

### Actual

Precedence is given to classifiers in [function _extractDeclaredLicense](https://github.com/clearlydefined/crawler/blob/master/providers/fetch/pypiFetch.js#L89-L94), which produces license `LGPL-2.0-only`.

```
  _extractDeclaredLicense(registryData) {
    const licenseFromClassifiers = this._extractLicenseFromClassifiers(registryData)
    if (licenseFromClassifiers) return licenseFromClassifiers
    const license = get(registryData, 'info.license')
    return license && spdxCorrect(license)
  }
```

qtomlinson:
@elrayle This will be a good topic to discuss in our next community meeting.  Historically, license was only parsed from classifiers.  This [commit](https://github.com/clearlydefined/crawler/commit/08bcf82b0cffbb5f8add833efa8f8e9ad318a069) adds the functionality to extract license from info.license.  To avoid breaking change, extracting from info.license is added as a fallback if no valid license is parsed from classifiers.

[...]

Issue #519, opened 2023-04-12
License set as Declared is not in the package


capfei:
https://clearlydefined.io/definitions/pypi/pypi/-/pillow/9.5.0

After harvesting `pypi/pypi/-/pillow/9.5.0` yesterday, it came back with `CAL-1.0` as the declared license but I could not find this anywhere in the harvested data.
<img width="842" alt="image" src="https://user-images.githubusercontent.com/5489711/231472267-56efc8b0-7ace-4906-8bb5-a63e8cb3bcf5.png">

I checked the previous version, 9.4.0, it looks like it also back as `CAL-1.0` in the harvested data.
https://clearlydefined.io/definitions/pypi/pypi/-/pillow/9.4.0

qtomlinson:
There are two problems identified by this issue:
1. extracting declared license in pypiFetch, 
- in ` _extractLicenseFromClassifiers`, `spdxCorrect("Historical Permission Notice and Disclaimer (HPND)")` returns `CAL-1.0`.
- if `info.license` (HPND) is used first instead, the resulting license from spdxCorrect("HPND") would be correct. 

This is related to the [discussion](https://github.com/clearlydefined/crawler/issues/523)
2. scancode summarizer (in service) should pick up the declared license detected in scancode raw data.

qtomlinson:
> 2. scancode summarizer (in service) should pick up the declared license detected in scancode raw data.

This has been fixed by [Commit](https://github.com/qtomlinson/service/commit/19e5a9a2f21972f582ed4706ba2a9a761c87b4a5).

capfei:
This is still happening: https://clearlydefined.io/definitions/pypi/pypi/-/flake8-isort/6.1.2/6.1.2

![Image](https://github.com/user-attachments/assets/b4153288-5e3a-4e37-a419-a28973648c76)

qtomlinson:
Refer to [registryData_flake8isort.json](https://github.com/user-attachments/files/20401447/registryData_flake8isort.json) for downloaded registry data.

The info.license field contains the full license text. When this text is parsed using spdx-correct, it is identified as CAL-1.0. Meanwhile, the license specified in the classifier is listed as 'GNU General Public License v2 (GPLv2)' which can be parsed to GPL-2.0-only. In this case, the classifier license is more accurate.

Issue #429, opened 2021-05-21
Incorrect declared license pulled by tooling


ariel11:
For the pillow packages, the tooling seems to think the declared license is CAL-1.0, even though the tooling identified the LICENSE as something else.  

Not sure where the crawler is pulling CAL-1.0 from but that is incorrect - e.g. https://clearlydefined.io/definitions/pypi/pypi/-/pillow/8.2.0

![image](https://user-images.githubusercontent.com/17935946/119176184-a03c7000-ba1f-11eb-84bb-9817e9b0cd9f.png)

ariel11:
This is still happening - https://clearlydefined.io/definitions/pypi/pypi/-/pillow/8.3.0

qtomlinson:
After syncing and rebuilding my local dev environment, response from http://localhost:4000/definitions/pypi/pypi/-/pillow/8.2.0  still showed license.declared as "CAL-1.0"

qtomlinson:
@ariel11 @capfei This seems to be a duplicate of https://github.com/clearlydefined/crawler/issues/519.  Could you please confirm? CAL-1.0 for https://clearlydefined.io/definitions/pypi/pypi/-/pillow/9.4.0.  Discussion on root cause, see https://github.com/clearlydefined/crawler/issues/519#issuecomment-1791404064

ariel11:
@qtomlinson - agree they look like the same issue.

ariel11:
This package has the same issue with "CAL-1.0" being the declared license finding, which is wrong - https://clearlydefined.io/definitions/pypi/pypi/-/matplotlib/3.9.2

qtomlinson:
Current [PR](https://github.com/clearlydefined/crawler/pull/586) on parsing precedence from info.license and info.classifier does not help this case:  [pypi/pypi/-/matplotlib/3.9.2](https://clearlydefined.io/definitions/pypi/pypi/-/matplotlib/3.9.2)
- from registryData: 
  - from the info.classifier, spdxCorrect('Python Software Foundation License') yield null
  - from the info.license, spdxCorrect yields 'CAL-1.0'

- from scancode, v30 yields NOASSERTION, may want to test v32.

PR #586, opened 2024-07-09
Derive license from info.license over classifiers in pypi registry data

qtomlinson:
Based on the [Python Packaging User Guide](https://packaging.python.org/en/latest/specifications/core-metadata/#license), it appears that the license field holds more specific information.  Change to parsing license field before the classifiers.  When classifier has the version info and the license field does not, use classifier first in that case.

Also updated the toolVersion in pypiExtract.

Task: #523 
Task: #519 
Task: #429


qtomlinson:
Here are the extract license information for 1000 [pypi components](https://github.com/user-attachments/files/16617404/licenses2.json).  The structure of each entry is as follows:
```
{
    "coordinates": "clearlydefined coordinates",
    "info.license": "info.license field from pypi registry data",
    "info.classifiers": "license extracted from info.classifiers field in pypi registry data"
}
```

qtomlinson:
Please find a more detailed JSON file with extracted SPDX licenses from the specified fields here: [licenses3.json](https://github.com/user-attachments/files/16626721/licenses3.json). In 150 instances, the non-empty SPDX licenses from the info.license field differ from those in the info.classifiers field. It appears that the licenses from the info.license field provide more detailed information.

Jeffrey-Luszcz:
I think we have a BSD vs BSD-2 vs BSD-3 clause problem we need to figure out how to handle. By reviewing the JSON file I see data with the following styles of problems:

License in repo is BSD-3 [ https://github.com/encode/starlette/blob/master/LICENSE.md ]but json says variously BSD & BSD-2:  
{
    "coordinates": "pypi/pypi/-/starlette/0.38.0",
    "info.classifiers": " BSD License",
    "info.license": null,
    "spdxFomClassifier": "BSD-2-Clause",
    "versionInClassifier": false,
    "versionInLicenseInfo": false
  },


List in repo is BSD-3 [ https://github.com/andialbrecht/sqlparse/blob/master/LICENSE ] but JSON says BSD or BSD-2
  {
    "coordinates": "pypi/pypi/-/sqlparse/0.5.1",
    "info.classifiers": " BSD License",
    "info.license": null,
    "spdxFomClassifier": "BSD-2-Clause",
    "versionInClassifier": false,
    "versionInLicenseInfo": false
  },

License in repo is BSD-3 [ https://github.com/google/re2/blob/main/LICENSE ]but json says variously BSD & BSD-2:  
  {
    "coordinates": "pypi/pypi/-/google-re2/1.1.20240702",
    "info.classifiers": " BSD License",
    "info.license": null,
    "spdxFomClassifier": "BSD-2-Clause",
    "versionInClassifier": false,
    "versionInLicenseInfo": false
  },

Jeffrey-Luszcz:
Also for LGPL we have a similar problem 

https://pypi.org/project/astroid/ License on pypi.org says "License: GNU Lesser General Public License v2 (LGPLv2) (LGPL-2.1-or-later)" but there's a mix of 2.0 and 2.1 and only and + in the json
 {
    "coordinates": "pypi/pypi/-/astroid/3.2.4",
    "info.classifiers": " GNU Lesser General Public License v2 (LGPLv2)",
    "info.license": "LGPL-2.1-or-later",
    "spdxFomClassifier": "LGPL-2.0-only",
    "versionInClassifier": true,
    "spdxFromLicenseInfo": "LGPL-2.1-or-later",
    "versionInLicenseInfo": true
  },

qtomlinson:
@Jeffrey-Luszcz This pull request is focused on extracting license information from the core metadata of the Python registry. This does not involve inspecting the license file, as that task is handled by other tools such as Licensee and ScanCode. The discrepancies you mentioned here are related to the cases where there is a mismatch between the registry metadata and the license file.

Jeffrey-Luszcz:
To close out this issue on my side. I support the change described (e.g. Derive license from info.license over classifiers)
@qtomlinson thanks for pulling together the json data, it really makes it clear.

My point in the comments above is that we should take another look at the mapping we have when we have a bare "BSD" license declared in these fields. We currently map it to BSD-2 and likely should map it to BSD-3 (the more common license choice) 
This is a separate issue and I'll open one for discussion.

elrayle (approved):
This all looks good.  Thanks for adding tests to capture the new possible conditions.

Issue #587, opened 2024-07-24
Not retrieving the latest version for some pod components


qtomlinson:
Expected;
When pod version is not specified, the most recent version should be retrieved

Observed:
0.9.1 is retrieved for xcbeautify. The most recent should be 0.17.0. See response from https://trunk.cocoapods.org/api/v1/pods/xcbeautify

Steps to reproduce:
Post to localhost:5000/requests with the following payload:

```
[
    {
        "type": "component",
        "url": "cd:/pod/cocoapods/-/xcbeautify"
    }
]
```

qtomlinson:
Thanks @yashkohli88 for reporting the issue!

PR #588, opened 2024-07-24
Fix fetching latest version for some pod components

qtomlinson:
Versions retrieved from the pod registry are sorted lexically. An example can be found at https://trunk.cocoapods.org/api/v1/pods/xcbeautify. Lexically, "0.17.0" < "0.9.1", so the latest version is not always the last item in the versions list. Change to sort by timestamp to retrieve the latest version.

Task: https://github.com/clearlydefined/crawler/issues/587


qtomlinson:
Thanks @yashkohli88 for reporting the issue!

elrayle (approved):
This is great.  Very nice bug fix.

PR #602, opened 2024-09-30
Deploy crawler to production azure

ljones140:
# What

Add's workflow that deploys a production crawler to Clearly Defineds Azure

**DO NOT MERGE:**

This is part of project to have Crawlers running on CD Azure as well as MSFT.

Currently the `cdcrawler-prod` that this workflow deploys is pointed at development azblob. To test deploy works.
After test we need to make additional changes so this workflow also publishes production image to Docker Hub as MSFT and GitHub are using.


jeffmendoza:
Is using OIDC a possibility for us to move away from credential secrets?
https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-azure

ljones140:
> Is using OIDC a possibility for us to move away from credential secrets? https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-azure

Yes that should work.

We log into Azure here in the base operations workflow https://github.com/clearlydefined/operations/blob/main/.github/workflows/app-deploy-to-azure.yml#L52-L55

Not sure if I can do as part of this task. But can create an issue on operations for this change.

ljones140:
Closing in favour of https://github.com/clearlydefined/crawler/pull/608

PR #605, opened 2024-10-18
Stop duplicate logging

ljones140:
Trailing a different approach.

Keep the same but invet the echo


ljones140:
Was not able to get working so closing

PR #611, opened 2024-10-23
Make scancode parallelism configurable

### LINKED ISSUE #609: Scancode parallelism is hardcoded to 2 processes
See here: https://github.com/clearlydefined/crawler/blob/0738316eaf5a9ef80f96cdaaeae38e2695688974/config/cdConfig.js#L98

Basically, scancode is always run at 2 parallel processes maximum, which becomes a bottleneck when running CD crawler on bigger machines.

The 2 makes sense as a default, since we see good CPU utilization in most of the current crawler deployments (which currently use at most 4 CPUs, as far as I know), but this number should be configurable for those special cases when larger instances are used for crawler.

RomanIakovlev:
Fixes #609

### DIFF
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -94,7 +94,7 @@ module.exports = {
         // '--quiet'
       ],
       timeout: 1000,
-      processes: 2,
+      processes: Number(process.env.CRAWLER_SCANCODE_PARALLELISM) || 2,
       format: '--json-pp'
     },
     source: {},


ljones140 (commented):
Great work this, hopefully this will improve processing speed.

Can you base it of master, when merged to master is should deploy to dev automatically.
And we can increase the scancode parallelism on there so the integration tests pass quicker

PR #612, opened 2024-10-23
Make scancode parallelism configurable

### LINKED ISSUE #609: Scancode parallelism is hardcoded to 2 processes

RomanIakovlev:
Fixes #609 
Replaces #611

### DIFF
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -95,7 +95,7 @@ module.exports = {
         // '--quiet'
       ],
       timeout: 1000,
-      processes: 2,
+      processes: Number(process.env.CRAWLER_SCANCODE_PARALLELISM) || 2,
       format: '--json-pp'
     },
     source: {},
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -95,7 +95,7 @@ module.exports = {
         // '--quiet'
       ],
       timeout: 1000,
-      processes: Number(process.env.CRAWLER_SCANCODE_PARALLELISM) || 2,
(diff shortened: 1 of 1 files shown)


ljones140 (commented):
Amazing work

ljones140 (approved):

config/cdConfig.js
```diff
@@ -95,7 +95,7 @@ module.exports = {
         // '--quiet'
       ],
       timeout: 1000,
-      processes: 2,
+      processes: Number(process.env.CRAWLER_SCANCODE_PARALLELISM) || 2,
```
elrayle, line 98: To allow this to be set using the normal configuration approach or from an env var, follow the pattern in... https://github.com/clearlydefined/crawler/blob/0738316eaf5a9ef80f96cdaaeae38e2695688974/ghcrawler/bin/www.js#L14

```suggestion
      processes: Number(config.get(CRAWLER_SCANCODE_PARALLELISM) || process.env.CRAWLER_SCANCODE_PARALLELISM) || 2,
```

elrayle (changes requested):
One change requested, then this looks good.

elrayle (approved):
Thanks for making this configurable.  This will really help with long term maintenance and scalability as needed.

PR #629, opened 2024-12-13
Upgrade Azure Storage SDK to a modern version

RomanIakovlev:
This change is a prerequisite to enable token-based authentication for the Azure Storage operations (blobs and queues). This change leaves the options to use the existing, connection string-based authentication though.


ghcrawler/providers/queuing/storageQueue.js
```diff
@@ -110,47 +89,21 @@ class StorageQueue {
...
-      this.client.getQueueMetadata(this.queueName, (result, error) => {
-        if (error) {
-          this.logger.error(error)
-          resolve(null)
-        }
-        resolve({ count: result[0].approximateMessageCount })
-      })
-    })
+    const properties = await this.queueClient.getProperties()
+    return { count: properties.approximateMessagesCount }
```
elrayle, line 106: This likely applies in other places as well.  In the original code, error handling logged the error.  In the new code, it appears that error handling depends on how `await` handles exceptions instead of using `try` and `catch` blocks to log the error in the same way as the original.

```suggestion
    try {
      const properties = await this.queueClient.getProperties()
      return { count: properties.approximateMessagesCount }
    } catch (error) {
      this.logger.error(error)  // Log the error
      return null  // Handle the error by returning null
    }
```

ghcrawler/providers/queuing/storageQueueManager.js
```diff
@@ -2,14 +2,31 @@
...
+        retryDelayInMs: 1000,
+        maxRetryDelayInMs: 120 * 1000,
+        tryTimeoutInMs: 30000,
+        retryPolicyType: StorageRetryPolicyType.EXPONENTIAL
+      }
+    }
+    if (connectionString) {
+      this.client = QueueServiceClient.fromConnectionString(connectionString, pipelineOptions)
+    } else {
+      this.client = new QueueServiceClient(
```
elrayle, line 24: This is a nice update here for maintaining the old approach and setting up a more flexible long term approach.

ghcrawler/providers/queuing/storageQueueManager.js
```diff
@@ -2,14 +2,31 @@
...
 const Request = require('../../lib/request')
 const StorageQueue = require('./storageQueue')
+const { DefaultAzureCredential } = require('@azure/identity')
 
 class StorageQueueManager {
-  constructor(connectionString) {
-    const retryOperations = new AzureStorage.ExponentialRetryPolicyFilter()
-    this.client = AzureStorage.createQueueService(connectionString).withFilter(retryOperations)
+  constructor(connectionString, options) {
+    const pipelineOptions = {
```
elrayle, line 12: I like the use of `pipelineOptions`.  Makes it easier to see the configs related to the queues.

ghcrawler/providers/storage/storageDocStore.js
```diff
@@ -43,133 +40,55 @@ class AzureStorageDocStore {
...
-        if (error) {
-          return reject(error)
-        }
-        resolve(true)
-      })
-    })
-  }
-
-  // This API can only be used for the 'deadletter' store because we cannot look up documents by type performantly
-  async count(type, force = false) {
```
elrayle, line 143: Can you talk about the reason this is deleted completely?

elrayle (commented):
A couple of comments.  I know you plan to do more testing, so I'll take another look once that is done.

ghcrawler/providers/storage/storageDocStore.js
RomanIakovlev: It was an oversight, I've brought it back. Thanks for catching it!

ghcrawler/providers/queuing/storageQueue.js
RomanIakovlev: Yeah it's a good point. There was one more place like this, I've updated it as well.

qtomlinson (commented):
Thanks for upgrading the sdk!  A couple of comments for you to consider.

ghcrawler/providers/storage/storageDocStore.js
```diff
@@ -150,24 +103,11 @@ class AzureStorageDocStore {
...
-        )
-      })
-      entryCount += result.entries.length
-    } while (continuationToken)
+    for await (const page of this.containerClient.listBlobsFlat().byPage({ maxPageSize: 1000 })) {
+      if (page.segment.blobItems) {
+        entryCount += page.segment.blobItems.length()
+      }
+    }
     memoryCache.put(key, entryCount, 60000)
```
qtomlinson, line 111: The constructing of this `key` needs to be adapted.  `this.name` is not longer available.  use `this.options.container` instead?

ghcrawler/providers/queuing/storageQueue.js
```diff
@@ -1,50 +1,46 @@
...
-            this.client.createMessage(this.queueName, body, option, (error, queueMessageResult) => {
-              if (error) {
-                return reject(error)
-              }
-              this._log('Queued', request)
-              resolve(this._buildMessageReceipt(queueMessageResult, request))
-            })
-          })
+          const queueMessageResult = await this.queueClient.sendMessage(body)
+          this._log('Queued', request)
```
qtomlinson, line 42: May want to consider encoding here. From [doc](https://learn.microsoft.com/en-us/javascript/api/@azure/storage-queue/queueclient?view=azure-node-latest#@azure-storage-queue-queueclient-sendmessage), the message content "must be in a format that can be included in an XML request with UTF-8 encoding" 

The legacy QueueService uses a default TextXmlQueueMessageEncoder; whereas the updated QueueService does not.

qtomlinson (commented):
Thanks for incorporating the changes.

qtomlinson:
Upon further testing, an error was observed when writing the `deadletter` to the blob store.

**Expected:** 
The error information should be persisted in the `deadletter` folder.

**Observed:** 
The error information was not persisted.

**Steps to reproduce:**
1. Set up Azure storage using Azurite. See [PR](https://github.com/clearlydefined/docker_dev_env_experiment/pull/20)
2. Trigger a harvest with an invalid type, e.g., a POST call to localhost:4000/harvest with the following payload:
    ```json
    [
        {
            "tool": "bogus",
            "coordinates": "pypi/pypi/-/platformdirs/4.2.0"
        }
    ]
    ```
    There should be an error because "bogus" is an invalid tool. The error information should be persisted in the `deadletter` folder. However, this error information is not persisted. The following error was observed:
    ```Error: options.metadata with value "1" must be of type string.```

qtomlinson:
Error writing deadletters can be addressed in a separate PR.

ljones140:
After I added a commit to change queue configuration I've deployed to dev and run the integration tests.

https://github.com/clearlydefined/operations/actions/runs/13264887396/job/37029783039#step:7:355

The failure looks ok to me as caused by a recent release to a package checked in the test

ljones140:
@qtomlinson I ran the Integration tests again and the queues cleared.
Screenshot 2025-02-12 at 15 00 40

ghcrawler/providers/storage/storageDocStore.js
ljones140: I will look into this

ljones140: addressed in 5d6eca2f12977cf00476f74f9fc506e18fbd41e9

qtomlinson (approved):
LGTM.  Please address https://github.com/clearlydefined/crawler/pull/629#issuecomment-2641969330 in a subsequent PR.

ljones140:
> LGTM. Please address [#629 (comment)](https://github.com/clearlydefined/crawler/pull/629#issuecomment-2641969330) in a subsequent PR.


@qtomlinson adding the fix to this PR as it was very simple edc8bb2


But did take a long time toe debug. Used the Azure docker env which was very helpful

ljones140:
Proof of deadletter working
Screenshot 2025-02-13 at 18 08 43

ghcrawler/lib/crawler.js
```diff
@@ -638,7 +638,7 @@ class Crawler {
       metadata.errorMessage = request._error.message
       metadata.errorStack = request._error.stack
     }
-    metadata.version = "1"
+    metadata.version = '1'
```
qtomlinson, line 641: Should this be updated to '1.1' as we are changing the file format for the deadletter here?

- Deadletters are internal exception information. The change here looks acceptable to me.
- Comparing the metadata structure of a deadletter and a harvested result (both are persisted documents in storage), there is no metadata.version in recent harvest results. Older harvest results prior to 2019 have metadata.version as a string. In the future, we may want to align with the metadata in harvested results and consider using schemaVersion instead.

qtomlinson (commented):
Thanks for the quick fix!  I am glad that the azurite setup makes debugging easier.

ghcrawler/lib/crawler.js
qtomlinson: @ljones140 we can merge this as is and discuss on the version upgrade during our regular sync?

Issue #630, opened 2025-01-28
Update Tool Version for Extract Functions in ClearlyDefined Processor


yashkohli88:
### Description
The tool version for the extract classes, which extend the ClearlyDefined processor, needs to be updated. This update is necessary to recompute/reharvest the components of package managers that utilize the `extractLicenseFromLicenseUrl` function to determine license. This function references `_licenseUrlOverrides`, which has been updated in [PR #1246](https://github.com/clearlydefined/service/pull/1246).

For further context and discussion, please refer to [this comment](https://github.com/clearlydefined/service/pull/1246#issuecomment-2533227880) within the mentioned PR.

### Steps to be followed
1. Identify the extract classes file that rely on `extractLicenseFromLicenseUrl`.
2. Ensure the patch  version is updated.

### Expected Outcome
- Components are re-harvested after the tool version is updated. New updated licenses are also getting reported in the re-harvested components.

PR #631, opened 2025-01-29
Update Tool Version for mavenExtract and nugetExtract

yashkohli88:
### Tasks Completed:

- [x] **Class Identification**: Identified the extract classes that depend on the `extractLicenseFromLicenseUrl` method, which utilizes the `_licenseUrlOverrides` list.
- [x] **Tool Version Updates**: Updated tool versions within the relevant extract classes for both Maven and NuGet.
- [x] **Test Case Addition**: Added a test case to validate and verify the updated versions, ensuring the re-harvesting is possible for coordinates of nuget and maven packages.
- [x] **Manual Re-harvesting**: Manually re-harvested Maven and NuGet components on a local system. This process confirmed that the ClearlyDefined tool processes these components as expected after updating the tool version, while other tools for these components are appropriately excluded.

This PR is related to https://github.com/clearlydefined/crawler/issues/630
Find the relevant comment and discussion here https://github.com/clearlydefined/service/pull/1246#issuecomment-2533227880

### DIFF
--- a/providers/process/mavenExtract.js
+++ b/providers/process/mavenExtract.js
@@ -13,7 +13,7 @@ class MavenExtract extends AbstractClearlyDefinedProcessor {
   }
 
   get toolVersion() {
-    return '1.3.0'
+    return '1.3.1'
   }
 
   canHandle(request) {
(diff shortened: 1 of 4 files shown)


elrayle (approved):
Thanks for the link the related PR to provide context.

PR #633, opened 2025-02-11
Separate crawler queue connection from harvest

ljones140:
This change is intended to give the ability to have a separate azblob connection to harvest blobs and crawler queues.

Organizations that run the crawler independently need to submit harvest results to CD Azure harvest but use their own Azure queues for crawler work queues.

By adding a boolean `isSpnAuth` to the `storageQueue` config it allows us to have crawler queues using a Azure SPN while having harvest blobs using a connection string. Without this it would have fallen back to the `az_blob.connection` which is the harvest results blob.

### DIFF
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -140,8 +140,9 @@ module.exports = {
       attenuation: {
         ttl: 3000
       },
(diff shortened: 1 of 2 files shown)


RomanIakovlev (approved):
Thanks! Looks good to me.

Issue #640, opened 2025-05-09
Make licensee's degree of parallelism configurable


jkbschmid:
Licensee is currently always limited to a maximum of 10 parallel processes per scan, which can lead to large fluctuations in the crawler's overall compute footprint, since it can suddenly start using 10 more cores.
https://github.com/clearlydefined/crawler/blob/29fd50a2c8abb39d3f40d8b4c819ac7cd2e1f0c0/providers/process/licensee.js#L55

Could we make this configurable, so users can reduce the degree of parallelism if they require a more predictable compute footprint?
This is similar to a previous issue about scancode's degree of parallelism (https://github.com/clearlydefined/crawler/issues/609).

PR #641, opened 2025-05-09
Make licensee's max degree of parallism configurable

jkbschmid:
via `CRAWLER_LICENSEE_PARALLELISM`.
The default of 10 stays as is.

Fixes https://github.com/clearlydefined/crawler/issues/640

### DIFF
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -22,7 +22,7 @@ const fetchedCacheTtlSeconds = config.get('CRAWLER_FETCHED_CACHE_TTL_SECONDS') |
 
 function getPositiveNum(configName, defaultValue) {
   const num = Number(config.get(configName))
-  return num > 0 ? num : defaultValue;
+  return num > 0 ? num : defaultValue
 }
 
 module.exports = {
@@ -73,7 +73,7 @@ module.exports = {
     gem: { githubToken },
     go: { githubToken },
     licensee: {
-      processes: getPositiveNum(CRAWLER_LICENSEE_PARALLELISM,10)
+      processes: getPositiveNum(CRAWLER_LICENSEE_PARALLELISM, 10)
     },
     maven: { githubToken },
     npm: { githubToken },
--- a/config/cdConfig.js
+++ b/config/cdConfig.js
@@ -73,7 +73,7 @@ module.exports = {
     gem: { githubToken },
     go: { githubToken },
     licensee: {
-      processes: getPositiveNum(CRAWLER_LICENSEE_PARALLELISM, 10)
+      processes: getPositiveNum('CRAWLER_LICENSEE_PARALLELISM', 10)
     },
     maven: { githubToken },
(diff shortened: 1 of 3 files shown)


config/cdConfig.js
```diff
@@ -67,7 +72,9 @@ module.exports = {
     },
     gem: { githubToken },
     go: { githubToken },
-    licensee: {},
+    licensee: {
+      processes: getLicenseeParallelism()
```
elrayle, line 76:
```suggestion
      processes: getPositiveNum(CRAWLER_LICENSEE_PARALLELISM,10)
```

config/cdConfig.js
```diff
@@ -20,6 +20,11 @@ const crawlerStoreProvider = config.get('CRAWLER_STORE_PROVIDER') || 'cd(file)'
 const maxRequeueAttemptCount = config.get('CRAWLER_MAX_REQUEUE_ATTEMPTS') || 5
 const fetchedCacheTtlSeconds = config.get('CRAWLER_FETCHED_CACHE_TTL_SECONDS') || 60 * 60 * 8 //8 hours
 
+function getLicenseeParallelism() {
+  const num = Number(config.get('CRAWLER_LICENSEE_PARALLELISM') || process.env.CRAWLER_LICENSEE_PARALLELISM)
+  return num > 0 ? num : 10;
+}
+
```
elrayle, line 27: I like that you added a validation that the config is a positive number.  It would be better to generalize this function so other configs that require a positive num can be refactored to use it when someone has time.  

```suggestion
function getPositiveNum(configName, defaultValue) {
  const num = Number(config.get(configName))
  return num > 0 ? num : defaultValue;
}

```

NOTE: I removed `process.env. CRAWLER_LICENSEE_PARALLELISM`.  None of the other configs come from env vars.  To introduce this approach would require the other configs to support env vars as well.  Can you say why you decided to go this route?

elrayle (changes requested):
Always good to provide more control to be able to tweak how the system runs and the resources it uses.  Thanks for submitting this!

I have a suggestion to ease long term maintenance.

config/cdConfig.js
jkbschmid: I added `process.env`, since it's also used for `CRAWLER_SCANCODE_PARALLELISM`.
That seems to be more of an outlier, so I think reading from `config` makes more sense.

elrayle (approved):
Thanks again for adding this config.  Looks good.

qtomlinson:
@jkbschmid @elrayle Thank you for the contribution!

Could you document this new addition in a manner similar to [this commit](https://github.com/clearlydefined/operations/commit/26f80173aa66cc1cd9dd0f1a18ae49dbdcc7aeef)?

jkbschmid:
@qtomlinson Thanks for the reminder, I opened https://github.com/clearlydefined/operations/pull/126.

PR #644, opened 2025-07-17
feat: update nodejs

elaine-mattos:


JamieMagee:
I compared this PR against current master to figure out what's still worth merging. Short version: about half of it landed through other work, but there are real gaps remaining.

What's already done separately: Node 24 is in CI and both Dockerfiles. The winston 3 migration in `logger.js` is complete, including dropping `winston-azure-application-insights`.

What's still broken on master that this PR fixes:

- REUSE is pinned at 3.0.1 in the Dockerfile. This PR bumps it to 5.0.2 — that hasn't happened anywhere else. (ScanCode did get bumped past this PR's target, to 32.5.0, so that part is moot.)
- Both Dockerfiles still use the legacy `ENV PORT 5000` syntax instead of `ENV PORT=5000`. Docker has been warning about this for a while.
- The DevDockerfile has a real bug: `ENV BUILD_NUMBER=$APP_VERSION` on line 8 references `APP_VERSION`, which isn't defined until line 9. The PR fixes the ordering.
- `scancode.js`, `licensee.js`, and `fsfeReuse.js` still call `logger.log()` for failures instead of `logger.error()` — that means errors don't get surfaced properly in Application Insights.
- No HTTPS proxy support in fetch.js. The `HttpsProxyAgent` addition is useful for anyone running behind a corporate proxy.
- Test fixtures still live under `32.1.0/` even though ScanCode is at 32.5.0 now.

Smaller stuff: .gitignore is missing a `.env` entry, package.json doesn't declare `engines.node`, and the configurable npm registry URL is an interesting idea but the PR has debug `console.log` statements that would need cleanup.

Honestly, given how much master has diverged, rebasing this into one coherent PR would be painful. It'd probably be easier to cherry-pick the remaining fixes into a few focused PRs: one for the Dockerfile fixes (REUSE bump, ENV syntax, DevDockerfile bug), one for the logging improvements, and maybe one for proxy support.

PR #656, opened 2025-11-11
Deprecate request/requestretry – Switch to Axios/axios-retry, Add Central getStream Helper

yashkohli88:
### Overview
This PR migrates away from the deprecated `request` and `requestretry` libraries in favor of an axios-based solution (`axios`, `axios-retry`). A new helper, `getStream`, is introduced to handle streaming downloads across package fetchers, reducing maintenance burden and improving long-term support.

### Main Changes

- **Refactor Download Logic**:  
  Replaces usage of `request` in all package fetchers (conda, debian, go, maven, npm, composer, pypi, ruby) with a new `getStream` helper in `lib/fetch.js`.
- **Retry Logic Update**:  
  Replaces `requestretry` with `axios-retry` for HTTP retry support. Centralizes retry logic in `callFetchWithRetry`.
- **Code Simplification & Maintenance**:  
  Removes now-unnecessary callback-based code and error handling, moving to promise-based, modern patterns.
- **Test Enhancements**:  
  Updates unit tests and stubs to use new axios-based methods and async/await.
- **API Consistency**:  
  All fetchers now consistently return streams and/or promise-based results.
- **Exponential Backoff Support**:  
  Adds exponential delay to retries.

### Notable File/Code Changes

- **lib/fetch.js**  
  - Adds `getStream` and `callFetchWithRetry`.
  - Switches HTTP requests and retry to `axios` + `axios-retry`.
- **All providers/fetch/*Fetch.js**  
  - Refactored from `request`/`requestretry` to the new helpers.
  - Error handling and stream logic updated to promises and async/await.
- **package.json**  
  - Removes `request` and `requestretry` dependencies.
- **Tests**  
  - Refactored all related test stubs and mocks to support new interfaces.

### Backwards Compatibility

- Code maintains the same external behavior for package downloads and retries; only the underlying HTTP/request implementation has changed.

### DIFF
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -29,10 +29,6 @@ function buildRequestOptions(request) {
   if (request.simple === false) {
     validateOptions.validateStatus = () => true
   }
-  if (request.gzip) {
-    if (!request.headers) request.headers = {}
-    request.headers['Accept-Encoding'] = 'gzip'
-  }
   return {
     method: request.method,
     url: request.url || request.uri,
@@ -93,11 +89,11 @@ async function callFetchWithRetry(url, options = {}, retryOptions = {}) {
     })
   }
 
-  options.resolveWithFullResponse ??= true
-  const request = { url, ...options }
+  const resolveWithFullResponse = options.resolveWithFullResponse ?? true
+  const request = { url, ...options, resolveWithFullResponse }
   try {
     const response = await callFetch(request, axiosInstance)
-    if (!options.resolveWithFullResponse) return response
+    if (!resolveWithFullResponse) return response
     return {
       statusCode: response.status,
       headers: response.headers,
@@ -105,7 +101,7 @@ async function callFetchWithRetry(url, options = {}, retryOptions = {}) {
       request: response.request
     }
   } catch (err) {
-    if (err.response) {
+    if (err.response && resolveWithFullResponse) {
       return {
         statusCode: err.response.status,
         headers: err.response.headers,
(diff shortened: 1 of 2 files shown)


pombredanne (approved):
Thanks. This looks great to me. Could you try to use an explicit value for the exponential delay?

lib/fetch.js
```diff
@@ -62,4 +66,38 @@ async function getStream(opt) {
   return await callFetch(request)
 }
 
-module.exports = { callFetch, withDefaults, defaultHeaders, getStream }
+async function callFetchWithRetry(url, options = {}, retryOptions = {}) {
+  const { maxAttempts = maximumAttempts, ...otherRetryOpts } = retryOptions
+  const axiosInstance = axios.create()
+  axiosRetry(axiosInstance, { retries: maxAttempts, retryDelay: exponentialDelay, ...otherRetryOpts })
```
JamieMagee: `requestretry`'s `maxAttempts: 3` meant 3 total attempts. `axios-retry`'s `retries: 3` means 3 retries *after* the initial attempt, so 4 total. This silently adds an extra attempt everywhere. The fetchTests.js: `maxAttempts: 5` makes 6 requests.

Should be `retries: maxAttempts - 1` to preserve the old behavior. Or rename the parameter to `retries` and update callers, but either way the current mapping is wrong.

providers/fetch/pypiFetch.js
JamieMagee: this still imports through requestRetryWithDefaults.js, which now just re-exports the entire `lib/fetch` module. Then it calls `requestRetry.callFetchWithRetry(...)` on pypiFetch.js instead of calling it as a function directly. Every other fetcher in this PR does:

```js
const { callFetchWithRetry: requestRetry } = require('../../lib/fetch')
```

After that, requestRetryWithDefaults.js has no remaining consumers and can be deleted.

